PR #16, opened 2024-08-22
Do not migrate if the router check does not pass

marcelofern:
Prior to this change, we weren't checking if the database router allowed the migration to proceed.

We thus add the router.allow_migrate() check, which is the same that the RunSQL operation runs internally. Ref:

- https://github.com/django/django/blob/7adb6dd98d50a238f3eca8c15b16b5aec12575fd/django/db/migrations/operations/special.py#L105

A possible error that would raise before this fix happens when the Django applicatoin has multiple databases and the migration should be routed to only apply on a particular database.

### DIFF
--- a/docs/usage/operations.rst
+++ b/docs/usage/operations.rst
@@ -6,7 +6,7 @@ Provides custom migration operations that help developers perform idempotent and
 Class Definitions
 -----------------
 
-.. py:class:: SaferAddIndexConcurrently(model_name: str, index: models.Index)
+.. py:class:: SaferAddIndexConcurrently(model_name: str, index: models.Index, hints: Any = None)
 
     Performs CREATE INDEX CONCURRENTLY IF NOT EXISTS without a lock_timeout
     value to guarantee the index creation won't be affected by any pre-set
@@ -16,6 +16,8 @@ Class Definitions
     :type model_name: str
     :param index: Any type of index supported by Django.
     :type index: models.Index
+    :param hints: Hints to be passed to the router as per https://docs.djangoproject.com/en/5.1/topics/db/multi-db/#hints
+    :type hints: Any
 
     **Why use this SaferAddIndexConcurrently operation?**
     -----------------------------------------------------
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -4,7 +4,7 @@ from textwrap import dedent
 from typing import Any
 
 from django.contrib.postgres import operations as psql_operations
-from django.db import migrations, models
+from django.db import migrations, models, router
 from django.db.backends.base import schema as base_schema
 from django.db.migrations.operations import base as migrations_base
 
@@ -49,10 +49,11 @@ class SaferAddIndexConcurrently(
 
     DROP_INDEX_QUERY = 'DROP INDEX CONCURRENTLY IF EXISTS "{}";'
 
-    def __init__(self, model_name: str, index: models.Index) -> None:
+    def __init__(self, model_name: str, index: models.Index, hints: Any = None) -> None:
         self.model_name = model_name
         self.index = index
         self.original_lock_timeout = ""
+        self.hints = {} if hints is None else hints
 
     def describe(self) -> str:
         return (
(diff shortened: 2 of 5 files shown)


docs/usage/operations.rst
```diff
@@ -6,7 +6,7 @@ Provides custom migration operations that help developers perform idempotent and
 Class Definitions
 -----------------
 
-.. py:class:: SaferAddIndexConcurrently(model_name: str, index: models.Index)
+.. py:class:: SaferAddIndexConcurrently(model_name: str, index: models.Index, hints: Any = None)
```
timb07, line 9: ❓  Should the type for `hints` be `Any` or `object`?

src/django_pg_migration_tools/operations.py
```diff
@@ -70,17 +71,22 @@ def database_forwards(
         to_state: migrations.state.ProjectState,
     ) -> None:
         self._ensure_not_in_transaction(schema_editor)
```
timb07, line 73: `_ensure_not_in_transaction` checks an attribute of the DB connection, which presumably establishes the connection if it hasn't been opened already. If we're not intending to apply this migration to this particular DB, then we should probably avoid opening the connection. In other words, push this check into the `if` block below.

src/django_pg_migration_tools/operations.py
```diff
@@ -90,14 +96,19 @@ def database_backwards(
...
```
timb07, line 98: ♻️

src/django_pg_migration_tools/operations.py
```diff
@@ -70,17 +71,22 @@ def database_forwards(
...
-        # Inject the IF NOT EXISTS because Django doesn't provide a handy
-        # if_not_exists: bool parameter for us to use.
-        index_sql = index_sql.replace(
-            "CREATE INDEX CONCURRENTLY", "CREATE INDEX CONCURRENTLY IF NOT EXISTS"
-        )
-        schema_editor.execute(index_sql)
-        self._ensure_original_lock_timeout_is_reset(schema_editor)
+        if router.allow_migrate(
+            schema_editor.connection.alias, app_label, **self.hints
+        ):
```
timb07, line 76: 🐼  Given the size of the indented block below, would it be clearer to do this instead?

```python
if not router.allow_migrate([...]):
    return
```

timb07 (approved):
LGTM overall; there are a couple of minor suggestions to consider.

docs/usage/operations.rst
marcelofern: I should've left a comment on the Pull Request for clarity, but essentially I am keeping the same type annotation as `django-stubs` for [consistency](https://github.com/typeddjango/django-stubs/blob/2dfa15bf94b038c4dae1590093981d3eacfa83e6/django-stubs/db/utils.pyi#L51).

src/django_pg_migration_tools/operations.py
```diff
@@ -49,10 +49,11 @@ class due to limitations of Django's AddIndexConcurrently operation.
 
     DROP_INDEX_QUERY = 'DROP INDEX CONCURRENTLY IF EXISTS "{}";'
 
-    def __init__(self, model_name: str, index: models.Index) -> None:
+    def __init__(self, model_name: str, index: models.Index, hints: Any = None) -> None:
         self.model_name = model_name
         self.index = index
         self.original_lock_timeout = ""
+        self.hints = hints or {}
```
delfick, line 56: 🐼 

```suggestion
        self.hints = {} if hints is None else hints
```

src/django_pg_migration_tools/operations.py
delfick: +1

src/django_pg_migration_tools/operations.py
marcelofern: That's a good observation. Django's `AddIndexConcurrently` behaves in the same way as in this PR. It runs the `_ensure_not_in_transaction` check [before verifying the router](https://github.com/django/django/blob/7adb6dd98d50a238f3eca8c15b16b5aec12575fd/django/contrib/postgres/operations.py#L136). That's why I decided to keep it here for consistency.

This seems to be a fail-early approach.

src/django_pg_migration_tools/operations.py
marcelofern: Replied here -> https://github.com/kraken-tech/django-pg-migration-tools/pull/16#discussion_r1728014628

src/django_pg_migration_tools/operations.py
marcelofern: Fair enough -> 6f30632a998687280364a47e82ac0290587642a6

src/django_pg_migration_tools/operations.py
marcelofern: Fair enough. The original does what [Django does](https://github.com/django/django/blob/7adb6dd98d50a238f3eca8c15b16b5aec12575fd/django/db/migrations/operations/special.py#L81), though I don't see harm on changing this.

Fixup: 77942bb7ba377deff5954cbb290fde9609108b80

delfick (approved):

PR #17, opened 2024-08-25
Workflow for tracking coverage

marcelofern:
# GitHub Workflow For Printing Coverage Result

This PR adds a new GitHub Workflow that uses the `coverage` package (from Pypi) to run reports on the testing coverage.

This code could be swapped for, or enhanced by, a [codecov subscription](https://about.codecov.io/) in future.

### DIFF
--- a/.github/workflows/coverage_comment.yaml
+++ b/.github/workflows/coverage_comment.yaml
@@ -1,0 +1,57 @@
+name: Comment Coverage Report
+
+on: pull_request
+
+env:
+  REPORT_PATH: /tmp/coverage_report
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 5
+    services:
+      postgres:
+        image: "postgres:16"
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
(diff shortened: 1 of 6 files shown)


.github/workflows/coverage_comment.yaml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          source .venv/bin/activate
+          make coverage_run
+          echo "## Coverage Report Results" > $REPORT_PATH
+          COVERAGE_FORMAT=markdown make --quiet coverage_report >> $REPORT_PATH
+
+      - name: Comment PR with coverage report
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          filePath: /tmp/coverage_report
```
marcelofern, line 56: Sadly I can't use $REPORT_PATH here as above. The `actions-comment-pull-request` Workflow doesn't seem to support env variables here.

.github/workflows/coverage_comment.yaml
```diff
@@ -0,0 +1,57 @@
...
+          comment_tag: pr_comment
```
marcelofern, line 57: This is so that instead of creating new comments for each push, it just creates one and updates it if a new push has happened.

delfick (approved):
Looks fine.

For papertrail, I asked outside of the PR why this couldn't be added on top of the existing tests and answer is speed. So the choice is to run code coverage on only one combination that the tests run under.

PR #18, opened 2024-08-26
SaferRemoveIndexConcurrently

marcelofern:
# Description

Adds a new Django migration operation class `SaferRemoveIndexConcurrently`. The name is self-explanatory, but here's a screenshot of the documentation (built with `make docs`) with more details (also available in the diff on the ante-penultimate commit).

image

### DIFF
--- a/docs/usage/operations.rst
+++ b/docs/usage/operations.rst
@@ -109,3 +109,97 @@ Class Definitions
                   index=models.Index(fields=["foo"], name="foo_idx"),
               ),
           ]
+
+
+
+.. py:class:: SaferRemoveIndexConcurrently(model_name: str, name: str)
+
+    Performs DROP INDEX CONCURRENTLY IF EXISTS without a lock_timeout
+    value to guarantee the index removal won't be affected by any pre-set
+    value of lock_timeout.
+
+    :param model_name: Model name in lowercase without underscores.
+    :type model_name: str
+    :param name: The name of the index to be deleted.
+    :type name: str
+
+    **Why use SaferRemoveIndexConcurrently?**
+    -----------------------------------------
+
+    Django already provides its own ``RemoveIndexConcurrently`` class, which is
+    available through *django.contrib.postres.operations*.
+
+    Django's operation performs a naive:
+
+    .. code-block:: sql
+
+      DROP INDEX CONCURRENTLY IF EXISTS <idx_name>;
+
+    Which has a few problems:
+
+    1. It might time out if an existing value of lock_timeout is pre-set.
+    2. If the operation started but failed because of a lock_timeout error,
+       the existing index won't be removed and it will be marked as INVALID.
+
+    Point 2. is usually not widely known. Even with the CONCURRENTLY
+    condition, the index removal needs to wait on locks, potentially being
+    interrupted by a lock_timeout thus leaving the existing index marked as
+    INVALID.
+
+    This custom ``SaferRemoveIndexConcurrently`` operation addresses theses
+    problems by running the following operations:
+
+    .. code-block:: sql
+
+      -- Necessary so that we can reset the value later.
+      SHOW lock_timeout;
+      -- Necessary to avoid lock timeouts. This is a safe operation as
+      -- DROP INDEX CONCURRENTLY does not lock concurrent selects, inserts,
+      -- updates, or deletes.
+      SET lock_timeout = 0;
+      -- Drop the index
+      DROP INDEX CONCURRENTLY IF EXISTS foo_idx;
+      -- Reset lock_timeout to its original value ("1s" as an example).
+      SET lock_timeout = '1s';
+
+    How to use
+    ----------
+
+    1. Remove the index from the relevant model:
+
+    .. code-block:: diff
+
+          class Meta:
+              indexes = (
+      -           # Existing index being removed.
+      -           models.Index(fields=["foo"], name="foo_idx"),
+                  # Another existing index not being removed.
+                  models.Index(fields=["bar"], name="bar_idx"),
+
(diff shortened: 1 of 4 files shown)


src/django_pg_migration_tools/operations.py
```diff
@@ -32,7 +32,7 @@ def safer_create_index(
         to_state: migrations.state.ProjectState,
         index: models.Index,
         model: type[models.Model],
-        operation: SaferAddIndexConcurrently,
+        operation: SaferAddIndexConcurrently | SaferRemoveIndexConcurrently,
```
delfick, line 35: can this instead be a Protocol class as the type you expect?

pyproject.toml
```diff
@@ -21,7 +21,7 @@ where = ["src"]
 
 [project]
 name = "django_pg_migration_tools"
-version = "0.1.2"
+version = "0.1.3"
```
delfick, line 24: I'd encourage the previous commit updating dependencies and the part of this commit that changes the version number be a separate PR dedicated to releasing a new version rather than bundling it with this change.

src/django_pg_migration_tools/operations.py
```diff
@@ -154,16 +116,116 @@ def _ensure_original_lock_timeout_is_reset(
             self.SET_LOCK_TIMEOUT_QUERY, {"lock_timeout": self.original_lock_timeout}
         )
 
-    # The following methods are necessary for Django to understand state
-    # changes.
-    def state_forwards(
-        self, app_label: str, state: migrations.state.ProjectState
+
+class SaferAddIndexConcurrently(
+    BaseIndexOperation, psql_operations.AddIndexConcurrently
```
delfick, line 121: is there anything stopping us from not using inheritence for composition? So that the SaferAddIndexConcurrently instance creates it's own instance of AddIndexConcurrently that it delegates to rather than being a subclass of AddIndexConcurrently?

src/django_pg_migration_tools/operations.py
marcelofern: There should be the only two classes that create/drop indexes, so I'm not sure there're gains in using structural typing here if we're not aiming for flexibility. On the side, nominal typing provides stronger contracts too.

pyproject.toml
marcelofern: Thanks. Are there any pros/cons to either approach I wouldn't be aware of? I thought that if a release is linked to a single change it would be best to keep it in the same PR so that it's easier to link to the scoped work from the tags  page. Similar to how some repos create a single release PR over multiple branches. Not particularly fussy about that either though.

src/django_pg_migration_tools/operations.py
marcelofern: I'd surely prefer composition if the code action was "AddIndexConcurrently, please go do something for me".

But in this case the code action is more like "AddIndexConcurrently, I need your interface so that Django's migration apparatus can understand my as an index creation operation".

This includes having methods that will register application database state alterations for example. But more importantly, I'm interested in picking up changes from upstream on that specific interface.

For example, a recent change in Django (new in 5.1) has added the "category" attribute to operations. In this case, `category = OperationCategory.ADDITION` (or REMOVAL), wouldn't be picked up by our classes unless we inherit from upstream. Otherwise, we'd have to keep different interfaces for different Django versions, which is less desirable.

pyproject.toml
delfick: yeah, certainly not a hard requirement, but my own sensibility is that releases should be separate as a general habit.

src/django_pg_migration_tools/operations.py
delfick: fair enough, inheritence sounds reasonable then.

delfick (approved):

pyproject.toml
marcelofern: [Here's](https://github.com/kraken-tech/django-pg-migration-tools/pull/19) a separated PR for the release. To be merged after this one is merged.

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -260,3 +260,170 @@ def test_when_not_allowed_to_migrate(self):
...
+
+        project_state = ProjectState()
+        project_state.add_model(ModelState.from_model(CharModel))
+        new_state = project_state.clone()
+
+        operation = operations.SaferRemoveIndexConcurrently(
+            "charmodel",
+            "char_field_idx",
+        )
+        # Proceed to try and add the index:
```
timb07, line 412: copy/paste: "remove" instead of "add"

docs/usage/operations.rst
```diff
@@ -109,3 +109,97 @@ Class Definitions
...
+      SET lock_timeout = 0;
+      -- Drop the index
+      DROP INDEX CONCURRENTLY IF EXISTS foo_idx;
+      -- Reset lock_timeout to its original value ("1s" as an example).
+      SET lock_timeout = '1s';
+
+    How to use
+    ----------
+
+    1. Remove the index to the relevant model:
```
timb07, line 168: copy/paste: should be "… from the relevant model"

timb07 (approved):
LGTM; just a couple of copy/paste typos to fix.

tests/django_pg_migration_tools/test_operations.py
marcelofern: Thx: 54d6cf9d4a37adffb3b0a4b70b7b301236f5cf93

docs/usage/operations.rst
marcelofern: Thx: 50d63778177336e3f3eeba70599eecf3543a7359

PR #20, opened 2024-09-03
Add a retry mechanism for `migrate_with_timeouts`

marcelofern:
This branch adds a retry mechanism for `migrate_with_timeouts`.

## Before
```bash
  ./manage.py migrate_with_timeouts \
    --lock-timeout-in-ms=10000 \
    --statement-timeout-in-ms=10000
  ```

## After

```bash
  ./manage.py migrate_with_timeouts \
    --lock-timeout-in-ms=10000 \
    --lock-timeout-max-retries=3 \
    --lock-timeout-retry-exp=2 \
    --lock-timeout-retry-min-wait-in-ms=3000 \
    --lock-timeout-retry-max-wait-in-ms=10000 \
    --statement-timeout-in-ms=10000 \
    --retry-callback--path="dotted.path.to.callback.function"
  ```

### DIFF
--- a/docs/usage/management_commands.rst
+++ b/docs/usage/management_commands.rst
@@ -1,27 +1,61 @@
 Management Commands
 ===================
 
-migrate_with_timeouts
----------------------
+``migrate_with_timeouts``
+-------------------------
 
 Runs database migrations with timeouts.
 
 This command wraps the :ref:`apply_timeouts() <apply_timeouts>` around
-Django's ``migrate`` command, exposing two extra arguments:
+Django's ``migrate`` command, exposing extra arguments to handle retries with
+backoff propagation.
 
-- ``--lock-timeout-in-ms``
-- ``--statement-timeout-in-ms``
++++++++++++++++++++++++
+lock_timeout arguments:
++++++++++++++++++++++++
 
- .. important::
+* ``--lock-timeout-in-ms``: Sets the lock_timeout value for the migration.
+* ``--lock-timeout-max-retries``: How many times to retry after getting a
+  lock_timeout. Defaults to zero, which means no retries.
+* ``--lock-timeout-retry-exp``: The exponential to use for exponential backoff
+  retry if retries are enabled. Defaults to 2.
+* ``--lock-timeout-retry-min-wait-in-ms``: minimum amount of time to wait
+  between lock_timeout retries in milliseconds. Defaults to 3s.
+* ``--lock-timeout-retry-max-wait-in-ms``: Maximum amount of time to wait
+  between lock_timeout retries in milliseconds. Defaults to 10s.
 
-   Both arguments are optional, but at least one must be provided!
+++++++++++++++++++++++++++++
+statement_timeout arguments:
+++++++++++++++++++++++++++++
 
-These arguments will set the value of Postgres' ``lock_timeout`` and
-``statement_timeout`` for the duration of the migration.
+* ``--statement-timeout-in-ms``: Sets statement_timeout for the migration.
 
-==========
+Note: Retry configurations are not available for statement timeouts at this
+stage. If you have a use case where statement retries would be useful please
+open an issue for discussion.
+
++++++++++++++++
+retry callback:
++++++++++++++++
+
+* ``--retry-callback-path``: Sets a callback to be called after a timeout
+  event has happened. An example is set below:
+
+  .. code-block:: python
+
+    import logging
+
+    from django_pg_migration_tools import timeouts
+    from django_pg_migration_tools.management.commands import migrate_with_timeouts
+
+
+    def timeout_callback(state: migrate_with_timeouts.RetryState) -> None:
+        logging.info("A lock timeout just happened!")
+        logging.info(f"{state.lock_timeouts_count} lock timeouts so far!")
+
+++++++++++
 How to use
-==========
+++++++++++
 
 1. Include ``django_pg_migration_tools`` in your ``INSTALLED_APPS``.
 
@@ -37,4 +71,11 @@ How to use
 
 .. code-block:: bash
 
-  ./manage.py migrate_with_timeouts --lock-timeout-in-ms=10000 --statement-timeout-in-ms=60000
+  ./manage.py migrate_with_timeouts \
+    --lock-timeout-in-ms=10000 \
+    --lock-timeout-max-retries=3 \
+    --lock-timeout-retry-exp=2 \
+    --lock-timeout-retry-min-wait-in-ms=3000 \
+    --lock-timeout-retry-max-wait-in-ms=10000 \
+    --statement-timeout-in-ms=10000 \
+    --retry-callback-path="dotted.path.to.callback.function"
--- a/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
+++ b/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
@@ -1,6 +1,8 @@
 import dataclasses
 import datetime
-from typing import Any
+import importlib
+import time
+from typing import Any, Protocol, cast
 
 from django.core.management import base
 from django.core.management.commands.migrate import Command as DjangoMigrationMC
@@ -30,25 +32,127 @@ class Command(DjangoMigrationMC):
             required=False,
             help="Value to set as statement_timeout in milliseconds.",
         )
+        parser.add_argument(
+            "--retry-callback-path",
+            dest="retry_callback_path",
+            type=str,
+            required=False,
+            help=(
+                "If retries are set, this argument can be used with the dotted path "
+                "to a function to be called between retries. The function must "
+                "follow this signature: f(retry_state) where `retry_state` "
+                "is the dataclass: "
+                "django_pg_migration_tools.management.commands.migrate_with_timeouts.RetryState."
+                "This callback can be useful for calling loggers after each "
+                "retry."
+            ),
+        )
+        parser.add_argument(
+            "--lock-timeout-max-retries",
+            dest="lock_timeout_max_retries",
+            type=int,
+            required=False,
+            default=0,
+            help=(
+                "How many times to retry after a lock timeout happened. "
+                "Defaults to zero, which means no retries - the migration "
+                "fails immediately upon lock timeout."
+            ),
+        )
+        parser.add_argument(
+            "--lock-timeout-retry-exp",
+            dest="lock_timeout_retry_exp",
+            type=int,
+            required=False,
+            default=2,
+            help="The value for the exponent for retry backoff delay. Defaults to 2.",
+        )
+        parser.add_argument(
+            "--lock-timeout-retry-max-wait-in-ms",
+            dest="lock_timeout_retry_max_wait_in_ms",
+            type=int,
+            required=False,
+            default=10_000,
+            help=(
+                "Sets a limit to the maximum length of time between subsequent "
+                "exponential backoff retries. Defaults to 10s."
+            ),
+        )
+        parser.add_argument(
+            "--lock-timeout-retry-min-wait-in-ms",
+            dest="lock_timeout_retry_min_wait_in_ms",
+            type=int,
+            required=False,
+            default=3_000,
+            help=(
+                "Sets a limit to the minimum length of time between subsequent "
+                "exponential backoff retries. Defaults to 3s."
+            ),
+        )
         super().add_arguments(parser)
 
     @base.no_translations
     def handle(self, *args: Any, **options: Any) -> None:
         timeout_options = MigrationTimeoutOptions.from_dictionary(options)
         timeout_options.validate()
+        retry_strategy = MigrateRetryStrategy(timeout_options=timeout_options)
 
-        with timeouts.apply_timeouts(
-            using=options["database"],
-            lock_timeout=timeout_options.lock_timeout,
-            statement_timeout=timeout_options.statement_timeout,
+        while retry_strategy.can_migrate():
+            try:
+                with timeouts.apply_timeouts(
+                    using=options["database"],
+                    lock_timeout=timeout_options.lock_timeout,
+                    statement_timeout=timeout_options.statement_timeout,
+                ):
+                    super().handle(*args, **options)
+                    return
+            except timeouts.DBLockTimeoutError as exc:
+                retry_strategy.increment_retry_count()
+                retry_strategy.wait()
+                retry_strategy.attempt_callback(exc)
+
+        raise base.CommandError(
+            f"Please consider trying a longer retry configuration or "
+            f"investigate whether there were long-running transactions "
+            f"during the migration. "
+            f"There were {retry_strategy.retries} lock timeouts. "
+            f"This happened because --lock-timeout-max-retries was set to "
+            f"{timeout_options.lock_retry_options.max_retries}."
+        )
+
+
+@dataclasses.dataclass
+class RetryState:
+    current_exception: timeouts.DBTimeoutError
+    lock_timeouts_count: int
+
+
+class RetryCallback(Protocol):
+    def __call__(self, retry_state: RetryState, /) -> None: ...  # pragma: no cover
+
+
+@dataclasses.dataclass(kw_only=True)
+class TimeoutRetryOptions:
+    max_retries: int
+    exp: int
+    max_wait: datetime.timedelta
+    min_wait: datetime.timedelta
+
+    def validate(self) -> None:
+        if (self.min_wait is not None and self.max_wait is not None) and (
+            self.min_wait > self.max_wait
         ):
-            super().handle(*args, **options)
+            raise ValueError(
+                "The minimum wait cannot be greater than the maximum wait for retries."
+            )
 
 
 @dataclasses.dataclass(frozen=True, kw_only=True)
 class MigrationTimeoutOptions:
     lock_timeout: datetime.timedelta | None
     statement_timeout: datetime.timedelta | None
+    lock_retry_options: TimeoutRetryOptions
+    retry_callback: RetryCallback | None
 
     @classmethod
     def from_dictionary(cls, options: dict[str, Any]) -> Self:
@@ -57,7 +161,24 @@ class MigrationTimeoutOptions:
                 options.pop("lock_timeout_in_ms", None)
             ),
             statement_timeout=_Parser.optional_positive_ms_to_timedelta(
-                options.pop("statement_timeout_in_ms", None)
+                options.pop("statement_timeout_in_ms", None),
+            ),
+            lock_retry_options=TimeoutRetryOptions(
+                max_retries=_Parser.required_positive_int(
+                    options.pop("lock_timeout_max_retries")
+                ),
+                exp=_Parser.required_positive_int(
+                    options.pop("lock_timeout_retry_exp")
+                ),
+                max_wait=_Parser.required_positive_ms_to_timedelta(
+                    options.pop("lock_timeout_retry_max_wait_in_ms")
+                ),
+                min_wait=_Parser.required_positive_ms_to_timedelta(
+                    options.pop("lock_timeout_retry_min_wait_in_ms")
+                ),
+            ),
+            retry_callback=_Parser.optional_retry_callback(
+                options.pop("retry_callback_path", None)
             ),
         )
 
@@ -67,6 +188,55 @@ class MigrationTimeoutOptions:
                 "At least one of --lock-timeout-in-ms or --statement-timeout-in-ms "
                 "must be specified."
             )
+        self.lock_retry_options.validate()
+
+
+class MigrateRetryStrategy:
+    timeout_options: MigrationTimeoutOptions
+    retries: int
+
+    def __init__(self, timeout_options: MigrationTimeoutOptions):
+        self.timeout_options = timeout_options
+        self.retries = 0
+
+    def wait(self) -> None:
+        exp = self.timeout_options.lock_retry_options.exp
+        min_wait = self.timeout_options.lock_retry_options.min_wait
+        max_wait = self.timeout_options.lock_retry_options.max_wait
+
+        if not self.can_migrate():
+            # No point waiting if we can't migrate.
+            return
+        try:
+            # self.retries is an integer, but it is turned into a float here
+            # because a huge exponentiation in Python between integers
+            # **never** overflows. Instead, the CPU is left trying to calculate
+            # the result forever and it will eventually return a memory error
+            # instead. Which we absolutely do not want. Please see:
+            # https://docs.python.org/3.12/library/exceptions.html#OverflowError
+            result = exp ** (float(self.retries))
+        except OverflowError:
+            result = max_wait.total_seconds()
+        wait = max(min_wait.total_seconds(), min(result, max_wait.total_seconds()))
+        time.sleep(wait)
+
+    def attempt_callback(self, current_exception: timeouts.DBTimeoutError) -> None:
+        if self.timeout_options.retry_callback:
+            self.timeout_options.retry_callback(
+                RetryState(
+                    current_exception=current_exception,
+                    lock_timeouts_count=self.retries,
+                )
+            )
+
+    def can_migrate(self) -> bool:
+        if self.retries == 0:
+            # This is the first time migration will run.
+            return True
+        return bool(self.retries <= self.timeout_options.lock_retry_options.max_retries)
+
+    def increment_retry_count(self) -> None:
+        self.retries += 1
 
 
 class _Parser:
@@ -88,3 +258,19 @@ class _Parser:
         if (not isinstance(value, int)) or (value < 0):
             raise ValueError(f"{value} is not a positive integer.")
         return value
+
+    @classmethod
+    def optional_retry_callback(cls, value: str | None) -> RetryCallback | None:
+        if not value:
+            return None
+
+        assert "." in value
+        module, attr_name = value.rsplit(".", 1)
+
+        # This raises ModuleNotFoundError, which gives a good explanation
+        # of the error already (see tests). We don't have to wrap this into
+        # our own exception.
+        callback_module = importlib.import_module(module)
+        callback = getattr(callback_module, attr_name)
+        assert callable(callback)
+        return cast(RetryCallback, callback)
--- a/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
+++ b/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
@@ -52,22 +52,13 @@ class MigrationTimeoutOptions:
 
     @classmethod
     def from_dictionary(cls, options: dict[str, Any]) -> Self:
-        statement_timeout_in_ms: int | None = options["statement_timeout_in_ms"]
-        lock_timeout_in_ms: int | None = options["lock_timeout_in_ms"]
-
-        statement_timeout: datetime.timedelta | None = None
-        if statement_timeout_in_ms is not None:
-            statement_timeout = datetime.timedelta(
-                seconds=int(statement_timeout_in_ms / 1_000)
-            )
-
-        lock_timeout: datetime.timedelta | None = None
-        if lock_timeout_in_ms is not None:
-            lock_timeout = datetime.timedelta(seconds=int(lock_timeout_in_ms / 1_000))
-
         return cls(
-            lock_timeout=lock_timeout,
-            statement_timeout=statement_timeout,
+            lock_timeout=_Parser.optional_positive_ms_to_timedelta(
+                options.pop("lock_timeout_in_ms", None)
+            ),
+            statement_timeout=_Parser.optional_positive_ms_to_timedelta(
+                options.pop("statement_timeout_in_ms", None)
+            ),
         )
 
     def validate(self) -> None:
@@ -76,3 +67,24 @@ class MigrationTimeoutOptions:
                 "At least one of --lock-timeout-in-ms or --statement-timeout-in-ms "
                 "must be specified."
             )
+
+
+class _Parser:
+    @classmethod
+    def optional_positive_ms_to_timedelta(
+        cls, value: int | None
+    ) -> datetime.timedelta | None:
+        if value is None:
+            return None
+        return cls.required_positive_ms_to_timedelta(value)
+
+    @classmethod
+    def required_positive_ms_to_timedelta(cls, value: int) -> datetime.timedelta:
+        value = cls.required_positive_int(value)
+        return datetime.timedelta(milliseconds=value)
+
+    @classmethod
+    def required_positive_int(cls, value: Any) -> int:
+        if (not isinstance(value, int)) or (value < 0):
+            raise ValueError(f"{value} is not a positive integer.")
+        return value
--- a/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
+++ b/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
@@ -1,8 +1,10 @@
+import dataclasses
 import datetime
 from typing import Any
 
 from django.core.management import base
 from django.core.management.commands.migrate import Command as DjangoMigrationMC
+from typing_extensions import Self
 
 from django_pg_migration_tools import timeouts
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
Comment (delfick) [inline src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py:1]: I typed up a suggestion, it would need comments I'm sure, and I haven't run mypy against it or tests, but as a starting point:

<details>

<summary>suggestion.py</summary>

```python
import dataclasses
import datetime
import functools
import importlib
import time
from typing import Any, Callable, Protocol, cast

from django.core.management import base
from django.core.management.commands.migrate import Command as DjangoMigrationMC
from django_pg_migration_tools import timeouts
from typing_extensions import Self


@dataclasses.dataclass(frozen=True, kw_only=True)
class RetryState:
    current_exception: timeouts.DBTimeoutError
    lock_timeouts_count: int
    statement_timeouts_count: int


class RetryCallback(Protocol):
    def __call__(self, retry_state: RetryState, /) -> None: ...


@dataclasses.dataclass(kw_only=True)
class TimeoutOptions:
    max_retries: int
    retry_exp: int
    retry_max_wait: datetime.timedelta
    retry_min_wait: datetime.timedelta

    timeouts: int = 0

    def can_retry(self) -> bool:
        return self.timeouts < self.max_retries

    def triggered(self) -> bool:
        return self.timeouts > 0

    def increment(self) -> None:
        self.timeouts += 1

    def wait(self, sleep: Callable[[float], None]) -> None:
        if not self.can_retry():
            # No point waiting if we can't retry.
            return

        try:
            # current_attempt is an integer, but it is turned into a float here
            # because a huge exponentiation in Python between integers
            # **never** overflows. Instead, the CPU is left trying to calculate
            # the result forever and it will eventually return a memory error
            # instead. Which we absolutely do not want. Please see:
            # https://docs.python.org/3.12/library/exceptions.html#OverflowError
            result = self.retry_exp ** (float(self.timeouts))
        except OverflowError:
            result = self.retry_max_wait.total_seconds()

        wait = max(
            self.retry_min_wait.total_seconds(),
            min(result, self.retry_max_wait.total_seconds()),
        )
        sleep(wait)

    def validate(self, cli_prefix: str) -> None:
        positive_fields = [
            self.max_retries,
            self.retry_exp,
            self.retry_max_wait,
            self.retry_min_wait,
        ]
        for f in positive_fields:
            # Validate for negative fields manually here instead of using a
            # custom argparse check as Django does not run the validator prior
            # calling handle(). At which point we don't have access to the
            # parser instance for calling argparse's validation.
            if isinstance(f, datetime.timedelta):
                # time fields are expressed in milisseconds and so are printed
                # back to the MC caller as such to avoid confusion.
                ms = int(f.total_seconds() * 1_000)
                if ms < 0:
                    raise ValueError(f"{ms} is not a positive value.")
            elif f < 0:
                raise ValueError(f"{f} is not a positive value.")

        if (self.retry_min_wait is not None and self.retry_max_wait is not None) and (
            self.retry_min_wait > self.retry_max_wait
        ):
            raise ValueError(
                "{cli_prefix}-retry-min-wait-in-ms cannot be greater than "
                "{cli_prefix}-retry-max-wait-in-ms."
            )


@dataclasses.dataclass(frozen=True, kw_only=True)
class MigrationOptions:
    overall_lock_timeout: datetime.timedelta | None
    overall_statement_timeout: datetime.timedelta | None

    lock_timeout: TimeoutOptions
    statement_timeout: TimeoutOptions
    retry_callback: Callable[[RetryState], None] | None

    def can_migrate(self) -> bool:
        if not self.lock_timeout.triggered() and not self.statement_timeout.triggered():
            # No timeouts implies this is the first time the migration will
            # run.
            return True

        return self.lock_timeout.can_retry() and self.statement_timeout.can_retry()

    def attempt_callback(self, current_exception: timeouts.DBTimeoutError) -> None:
        if self.retry_callback is None:
            return

        self.retry_callback(
            RetryState(
                current_exception=current_exception,
                lock_timeouts_count=self.lock_timeouts,
                statement_timeouts_count=self.statement_timeouts,
            )
        )

    @classmethod
    def from_dictionary(cls, options: dict[str, object]) -> Self:
        return cls(
            overall_lock_timeout=_Parser.optional_msg_to_timedelta(
                options,
                "lock_timeout_in_ms",
            ),
            overall_statement_timeout=_Parser.optional_ms_to_timedelta(
                options,
                "statement_timeout_in_ms",
            ),
            lock_timeout=TimeoutOptions(
                max_retries=_Parser.required_int(
                    options,
                    "lock_timeout_max_retries",
                ),
                retry_exp=_Parser.required_int(
                    options,
                    "lock_timeout_retry_exp",
                ),
                retry_max_wait=_Parser.required_ms_to_timedelta(
                    options,
                    "lock_timeout_retry_max_wait_in_ms",
                ),
                retry_min_wait=_Parser.required_ms_to_timedelta(
                    options,
                    "lock_timeout_retry_min_wait_in_ms",
                ),
            ),
            statement_timeout=TimeoutOptions(
                max_retries=_Parser.required_int(
                    options,
                    "statement_timeout_max_retries",
                ),
                retry_exp=_Parser.required_int(
                    options,
                    "statement_timeout_retry_exp",
                ),
                retry_max_wait=_Parser.required_ms_to_timedelta(
                    options,
                    "statement_timeout_retry_max_wait_in_ms",
                ),
                retry_min_wait=_Parser.required_ms_to_timedelta(
                    options,
                    "statement_timeout_retry_min_wait_in_ms",
                ),
            ),
            retry_callback=_Parser.optional_retry_callback(
                options,
                "retry_callback_path",
            ),
        )

    def validate(self) -> None:
        if self.overall_statement_timeout is None and self.overall_lock_timeout is None:
            raise ValueError(
                "At least one of --lock-timeout-in-ms or --statement-timeout-in-ms "
                "must be specified."
            )

        self.lock_timeout.validate(cli_prefix="--lock-timeout")
        self.statement_timeout.validate(cli_prefix="--statement-timeout")


@dataclasses.dataclass(frozen=True, kw_only=True)
class Migrator:
    options: MigrationOptions

    def migrate(self, using: str, handle: Callable[[], None]) -> None:
        self.options.validate()

        migration_applied: bool = False
        while not migration_applied and self.options.can_migrate():
            try:
                with timeouts.apply_timeouts(
                    using=using,
                    lock_timeout=self.options.lock_timeout,
                    statement_timeout=self.options.statement_timeout,
                ):
                    handle()
                    migration_applied = True
            except timeouts.DBLockTimeoutError as exc:
                self.options.lock_timeout.increment()
                self.options.lock_timeout.wait(time.sleep)
                self.options.attempt_callback(exc)
            except timeouts.DBStatementTimeoutError as exc:
                self.options.statement_timeout.increment()
                self.options.statement_timeout.wait(time.sleep)
                self.options.attempt_callback(exc)

        if migration_applied is False:
            raise base.CommandError(
                f"Please consider trying a longer retry configuration or "
                f"investigate whether there were long-running transactions "
                f"during the migration. "
                f"There were {self.options.lock_timeouts} lock timeouts and "
                f"{self.options.statement_timeouts} statement timeouts. This happened "
                f"because --lock-timeout-max-retries was set to "
                f"{self.options.lock_timeout_max_retries} and "
                f"--statement-timeout-max-retries was set to "
                f"{self.options.statement_timeout_max_retries}."
            )


class Command(DjangoMigrationMC):
    help = (
        "Wrapper around Django's migrate command that sets a lock_timeout "
        "value to ensure migrations don't wait for locks for too long."
    )

    def add_arguments(self, parser: base.CommandParser) -> None:
        parser.add_argument(
            "--lock-timeout-in-ms",
            dest="lock_timeout_in_ms",
            type=int,
            required=False,
            help="Value to set as lock_timeout in milliseconds.",
        )
        parser.add_argument(
            "--statement-timeout-in-ms",
            dest="statement_timeout_in_ms",
            type=int,
            required=False,
            help="Value to set as statement_timeout in milliseconds.",
        )
        parser.add_argument(
            "--retry-callback--path",
            dest="retry_callback_path",
            type=str,
            required=False,
            help=(
                "If retries are set, this argument can be used with the dotted path "
                "to a function to be called between retries. The function must "
                "follow this signature: f(retry_state) where `retry_state` "
                "is the dataclass: "
                "django_pg_migration_tools.management.commands.migrate_with_timeouts.RetryState."
                "This callback can be useful for calling loggers after each "
                "retry."
            ),
        )
        parser.add_argument(
            "--lock-timeout-max-retries",
            dest="lock_timeout_max_retries",
            type=int,
            required=False,
            default=0,
            help=(
                "How many times to retry after a lock timeout happened. "
                "Defaults to zero, which means no retries - the migration "
                "fails immediatelly upon lock timeout."
            ),
        )
        parser.add_argument(
            "--lock-timeout-retry-exp",
            dest="lock_timeout_retry_exp",
            type=int,
            required=False,
            default=2,
            help=(
                "The value for the exponent for retry backoff propagation. "
                "Defaults to 2."
            ),
        )
        parser.add_argument(
            "--lock-timeout-retry-max-wait-in-ms",
            dest="lock_timeout_retry_max_wait_in_ms",
            type=int,
            required=False,
            default=10_000,
            help=(
                "Sets a limit to the maximum length of time between subsequent "
                "exponential backoff propagation retries. Defaults to 10s."
            ),
        )
        parser.add_argument(
            "--lock-timeout-retry-min-wait-in-ms",
            dest="lock_timeout_retry_min_wait_in_ms",
            type=int,
            required=False,
            default=3_000,
            help=(
                "Sets a limit to the minimum length of time between subsequent "
                "exponential backoff propagation retries. Defaults to 3s."
            ),
        )
        parser.add_argument(
            "--statement-timeout-max-retries",
            dest="statement_timeout_max_retries",
            type=int,
            required=False,
            default=0,
            help=(
                "How many times to retry after a statement timeout happened. "
                "Defaults to zero, which means no retries - the migration "
                "fails immediatelly upon statement timeout."
            ),
        )
        parser.add_argument(
            "--statement-timeout-retry-exp",
            dest="statement_timeout_retry_exp",
            type=int,
            required=False,
            default=2,
            help=(
                "The value for the exponent for retry backoff propagation. "
                "Defaults to 2."
            ),
        )
        parser.add_argument(
            "--statement-timeout-retry-max-wait-in-ms",
            dest="statement_timeout_retry_max_wait_in_ms",
            type=int,
            required=False,
            default=10_000,
            help=(
                "Sets a limit to the length of time between subsequent "
                "exponential backoff propagation retries. Defaults to 10s."
            ),
        )
        parser.add_argument(
            "--statement-timeout-retry-min-wait-in-ms",
            dest="statement_timeout_retry_min_wait_in_ms",
            type=int,
            required=False,
            default=3_000,
            help=(
                "Sets a limit to the minimum length of time between subsequent "
                "exponential backoff propagation retries. Defaults to 3s."
            ),
        )
        super().add_arguments(parser)

    @base.no_translations
    def handle(self, *args: Any, **options: Any) -> None:
        migration_options = MigrationOptions.from_dictionary(options)
        database: str = options["database"]
        handle = functools.partial(super().handle, *args, **options)
        Migrator(options=migration_options).migrate(using=database, handle=handle)


class _Parser:
    @classmethod
    def optional_ms_to_timedelta(
        cls, options: dict[str, object], key: str
    ) -> datetime.timedelta | None:
        value = options.get(key)
        if value is None:
            return None

        return cls.required_ms_to_timedelta(options, key)

    @classmethod
    def required_ms_to_timedelta(
        cls, options: dict[str, object], key: str
    ) -> datetime.timedelta | None:
        value = options.pop(key)
        assert isinstance(value, int)
        return cls.ms_to_timedelta(value)

    @classmethod
    def required_int(cls, options: dict[str, object], key: str) -> int:
        value = options.pop(key)
        assert isinstance(value, int)
        return value

    @classmethod
    def optional_retry_callback(
        cls, options: dict[str, object], key: str
    ) -> RetryCallback | None:
        value = options.get(key)
        if value is None:
            return None
        return cls.required_retry_casllback(options, key)

    @classmethod
    def required_retry_callback(
        cls, options: dict[str, object], key: str
    ) -> RetryCallback:
        value = options.pop(key)

        assert "." in value
        module, attr_name = value.rsplit(".", 1)

        # This raises ModuleNotFoundError, which gives a good explanation
        # of the error already (see tests). We don't have to wrap this into
        # our own exception.
        callback_module = importlib.import_module(module)
        callback = getattr(callback_module, attr_name)
        assert callable(callback)

        # probs unnecessasry to do inspect.signature introspection
        return cast(RetryCallback, callback)

    @classmethod
    def ms_to_timedelta(value: int) -> datetime.timedelta:
        return datetime.timedelta(seconds=int(value / 1_000))
```
</details>

means that we aren't setting variables on the management command outside of it's constructor and we split the cli parsing from the logic

```
    @base.no_translations
    def handle(self, *args: Any, **options: Any) -> None:
        migration_options = MigrationOptions.from_dictionary(options)
        database: str = options["database"]
        handle = functools.partial(super().handle, *args, **options)
        Migrator(options=migration_options).migrate(using=database, handle=handle)
```

And also makes it so that the timeout options are their own object so that the lock timeout and the statement timeout are two different objects rather than duplicated code

I also think it makes the migrate command itself easier to read too

delfick (commented):
nice :)

I've added a suggestion that moves the code around a little.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -28,32 +54,304 @@ def add_arguments(self, parser: base.CommandParser) -> None:
...
 
-        with timeouts.apply_timeouts(
-            using=options["database"],
-            lock_timeout=lock_timeout,
-            statement_timeout=statement_timeout,
-        ):
-            super().handle(*args, **options)
+    def _migrate(self, *django_migrate_args: Any, **django_migrate_kwargs: Any) -> None:
+        migration_applied: bool = False
+        while self._can_migrate() and (migration_applied is False):
```
delfick, line 260:
```suggestion
        while self._can_migrate():
```
and then do a `return` instead of `migration_applied = True`

And then the raising of the exception does need the `if not migration_applied` guard cause the only time that would be reached would be if the migration didn't happen.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -28,32 +54,304 @@ def add_arguments(self, parser: base.CommandParser) -> None:
...
+        )
+        self.lock_timeout_retry_min_wait = _parse_int_in_ms_to_timedelta(
+            options.pop("lock_timeout_retry_min_wait_in_ms")
+        )
+        self.retry_callback = None
+        if callback_path := options.pop("retry_callback_path", None):
+            # This raises ModuleNotFoundError, which gives a good explanation
+            # of the error already (see tests). We don't have to wrap this into
+            # our own exception.
+            *module, callback = callback_path.split(".")
```
delfick, line 202: in my suggestion I do a `module, attr_name = callaback_path.rsplit(".")` here to avoid needing the ".".join(module). I also make sure the path has at least one dot in it and I `assert callable(callback)`

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -69,3 +61,13 @@ def _migrate(self, *django_migrate_args: Any, **django_migrate_kwargs: Any) -> N
             statement_timeout=self.statement_timeout,
         ):
             super().handle(*django_migrate_args, **django_migrate_kwargs)
+
+
+def _parse_non_required_ms_option_to_timedelta(
+    options: Any, key: str
+) -> datetime.timedelta | None:
+    value = options.pop(key, None)
```
timb07, line 349: 🐼  The word "parse" in the function name suggests only parsing, not mutating the `options` dict. Naming things is hard, but perhaps "pop" or "extract" would be better.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -1,15 +1,38 @@
...
+
 class Command(DjangoMigrationMC):
+    retry_callback: Callable[[RetryState], None] | None
+
     statement_timeout: datetime.timedelta | None
+    statement_timeouts: int
+    statement_timeouts_max_retries: int
+    statement_timeout_retry_exp: int
+    statement_timeout_retry_max_wait: datetime.timedelta
+    statement_timeout_retry_min_wait: datetime.timedelta
```
timb07, line 28: 🤔  There's a strong use case for having retries for lock timeouts, but I'm struggling to see why retrying in case of statement timeouts would be useful. Do we expect a statement that times out the first time to somehow become faster and not to timeout when retried? What's an example of such a statement where that's likely?

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -31,6 +54,111 @@ def add_arguments(self, parser: base.CommandParser) -> None:
             required=False,
             help="Value to set as statement_timeout in milliseconds.",
         )
+        parser.add_argument(
+            "--retry-callback--path",
```
timb07, line 58: Extra hyphen: `...callback--path`

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -31,6 +54,111 @@ def add_arguments(self, parser: base.CommandParser) -> None:
...
+        parser.add_argument(
+            "--lock-timeout-max-retries",
+            dest="lock_timeout_max_retries",
+            type=int,
+            required=False,
+            default=0,
+            help=(
+                "How many times to retry after a lock timeout happened. "
+                "Defaults to zero, which means no retries - the migration "
+                "fails immediatelly upon lock timeout."
```
timb07, line 81: typo: "immediately"

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -31,6 +54,111 @@ def add_arguments(self, parser: base.CommandParser) -> None:
...
+            ),
+        )
+        parser.add_argument(
+            "--lock-timeout-retry-exp",
+            dest="lock_timeout_retry_exp",
+            type=int,
+            required=False,
+            default=2,
+            help=(
+                "The value for the exponent for retry backoff propagation. "
```
timb07, line 91: 🐼  "propagation" doesn't seem like the right word, but it's not clear to me what the right word is. Would "delay" work?

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -69,3 +61,13 @@ def _migrate(self, *django_migrate_args: Any, **django_migrate_kwargs: Any) -> N
...
+
+
+def _parse_non_required_ms_option_to_timedelta(
+    options: Any, key: str
+) -> datetime.timedelta | None:
+    value = options.pop(key, None)
+    if value is None:
+        return None
+    assert isinstance(value, int)
+    return datetime.timedelta(seconds=int(value / 1_000))
```
timb07, line 357: Why have the argument in milliseconds if it's going to be rounded down to a whole number of seconds anyway?

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -54,13 +210,137 @@ def _validate(self) -> None:
...
+                self.statement_timeouts += 1
+                self._wait(
+                    exp=self.statement_timeout_retry_exp,
+                    min_wait=self.statement_timeout_retry_min_wait,
+                    max_wait=self.statement_timeout_retry_max_wait,
+                    current_attempt=self.statement_timeouts,
+                )
+                self._attempt_callback(exc)
+
+        if migration_applied is False:
```
timb07, line 288: Pet peeve: comparison of booleans.

Prefer: `if not migration_applied:`

(and also above)

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -54,13 +210,137 @@ def _validate(self) -> None:
...
+            # No point waiting if we can't migrate.
+            return
+        try:
+            # current_attempt is an integer, but it is turned into a float here
+            # because a huge exponentiation in Python between integers
+            # **never** overflows. Instead, the CPU is left trying to calculate
+            # the result forever and it will eventually return a memory error
+            # instead. Which we absolutely do not want. Please see:
+            # https://docs.python.org/3.12/library/exceptions.html#OverflowError
+            result = exp ** (float(current_attempt))
```
timb07, line 329: How many retries do we expect users to attempt? More than thousands???

Counterpoint: this code will cope with any number of retries, and with no particular downsides apart from the need to explain why the cast to float is done.

docs/usage/management_commands.rst
```diff
@@ -1,27 +1,70 @@
...
+  between statement timeout retries in milliseconds. Defaults to 3s.
+* ``--statement-timeout-retry-max-wait-in-ms``: Maximum amount of time to wait
+  between statement timeout retries in milliseconds. Defaults to 10s.
 
-==========
++++++++++++++++
+retry callback:
++++++++++++++++
+
+* ``--retry-callback--path``: Sets a callback to be called after a timeout
```
timb07, line 45: copy-paste: remove double hyphen here: `...callback--path`

docs/usage/management_commands.rst
```diff
@@ -37,4 +80,15 @@ How to use
...
+    --lock-timeout-max-retries=3 \
+    --lock-timeout-retry-exp=2 \
+    --lock-timeout-retry-min-wait-in-ms=3000 \
+    --lock-timeout-retry-max-wait-in-ms=10000 \
+    --statement-timeout-in-ms=10000 \
+    --statement-timeout-max-retries=3 \
+    --statement-timeout-retry-exp=2 \
+    --statement-timeout-retry-min-wait-in-ms=3000 \
+    --statement-timeout-retry-max-wait-in-ms=10000 \
+    --retry-callback--path="dotted.path.to.callback.function"
```
timb07, line 94: ♻️

timb07 (commented):
I'm unsure of the need for retries on statement timeout — please convince me that the extra complexity they add is worthwhile.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: Thanks for the detailed reply @delfick.

I'm happy to make the code more OOP. There are a few things I'd like to discuss first so that we can come to the best implementation:

- If we're breaking down the implementation into classes, I think the `_Parser` would be the best place to raise the basic input validation errors. E.g.  "is this a positive integer?", "is this value larger than that value?". Given that these sound like "interface-level" validation, it would mean that `TimeoutOptions` and `MigrationOptions` would not have to validate their data as they'd be instantiated with valid data already (from the parser). This is in the same vain as Django's `handle()` "parser" argument which does validation on the interface level. Let me know what you think.

- The `Migrator` class logic seems small enough to inline within the `Command` itself. What do you think? Though I don't have strong preferences I find it more readable if the "procedural code is close to where the procedure happens" when there is no need for a particular piece of code to be reusable.

I will take the other suggestions as I agree that my "dumbest-possible solution" can be improved with a few extra dataclasses for the retry bit and the protocol.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: @timb07 

> Do we expect a statement that times out the first time to somehow become faster and not to timeout when retried?

Good question. I think that this is a possible situation because of temporary load spikes on the database, caused by either background tasks or maintenance tasks starving server resources.

E.g., a vacuum finishing between retries could make the schema change query execute faster and thus pass on a second attempt. Adding a unique constraint to a table with many dead rows is an example of such query. The constraint generation would need to step through many dead-rows and thus have to load more pages from disk to finish its job before the vacuum which could substantially change the results.

I do agree though, that it doesn't feel as useful as lock_timeout retries and its usage might be limited to fewer situations. 

It almost makes sense to add a exponential back-off for setting `statement_timeout` to ever increasing values up to a limit, but that sounds like a different feature.

timb07: I agree that the examples you've given do fit the scenario I asked about. But they're not terribly convincing as a justification for the need for this feature. Whereas lock timeouts + automatic retries is well-established best practice (e.g. [this blogpost](https://postgres.ai/blog/20210923-zero-downtime-postgres-schema-migrations-lock-timeout-and-retries#graceful-schema-changes-lock_timeout-and-retries)) for applying migrations. 

And let's go back to why we might want a statement timeout for migrations in the first place. The main reason (as I understand it) is that we don't want migrations to take too long, because they delay the overall deployment process. So if we hit the statement timeout, isn't retrying the wrong thing to do?

marcelofern: No problems, I will remove this part of the feature and if we find better use cases in the future we can add it back.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
delfick: the argument for _Parser and Migrator being separate to the argparse is that the management command is one entry point to the migration functionality, as opposed to us making it only entry point (even if the current intention is that it'll always be the only entry point). Like, the duplication is coincidental only.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: I've picked up the `pop` prefix instead.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: Addressed in:

- bd5a5ed87fc11223dcef9f096e6e6e31cb5a48d8
- 5c35e23041229870259680286df1b2512d051894

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: Fat finger + auto complete gone wrong:
- 5f2326fb9178287c024f25f87dbfec61b03e8f08
- 79e7b8fe75c521a8bfeb4be21dbe7564ebf20353

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: cac2c606b13ddb6d57fa1d14cb861ce21ef8281f

marcelofern: cac2c606b13ddb6d57fa1d14cb861ce21ef8281f

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: 03d3ed2613043d4dc358fc8d7e368d515dafb016

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: [Addressed] That's a problem. I wasn't aware that `timedelta` had a `miliseconds` argument as I was biased by the author's preference for microseconds:

image

docs/usage/management_commands.rst
marcelofern: Refer to https://github.com/kraken-tech/django-pg-migration-tools/pull/20#discussion_r1760473678

docs/usage/management_commands.rst
marcelofern: Refer to https://github.com/kraken-tech/django-pg-migration-tools/pull/20#discussion_r1760473678

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: 2449e41f0a76fcfd8f2a7fc69ba531c7c4d6571d

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: Not in the thousands. Hopefully not even in the hundreds.
You can, of course, run into overflow errors in less than a thousand attempts for a high exponential value - but I also hope this to not be a common scenario. Giving that I am "hoping for many things" I agree that it's best to make the code work for any number of retries lest an arbitrary limiting number is incorrectly chosen and we have to go through a soft deprecation process.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: @delfick I've taken 80% of your suggestions onboard. With the exception of the Migrator class (reason above) and a few modifications to the original.

- Fixup: 19fc7cbc8b4aaf79a472fc2f5601545de97f868b

I'm not sure I understood your reply above. Happy to discuss further if you are not content with the fixup above.

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -1,13 +1,38 @@
...
+
+    statement_timeout: datetime.timedelta | None
+
+    lock_timeout: datetime.timedelta | None
+    lock_timeouts: int
+    lock_timeouts_max_retries: int
+    lock_timeout_retry_exp: int
+    lock_timeout_retry_max_wait: datetime.timedelta
+    lock_timeout_retry_min_wait: datetime.timedelta
+
```
delfick, line 35: these aren't needed anymore right?

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -28,32 +53,237 @@ def add_arguments(self, parser: base.CommandParser) -> None:
...
+        migration_applied: bool = False
+        while retry_strategy.can_migrate() and (not migration_applied):
+            try:
+                with timeouts.apply_timeouts(
+                    using=options["database"],
+                    lock_timeout=timeout_options.lock_timeout,
+                    statement_timeout=timeout_options.statement_timeout,
+                ):
+                    super().handle(*args, **options)
+                    migration_applied = True
```
delfick, line 130:
```suggestion
                    return
```

And then remove the `migration_applied` flag

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
delfick: I stand by making the actual logic for the migration be interface agnostic and the django management command nothing more than a way of getting input from the cli into that migration logic, but otherwise I like what you've done here.

Are you able to remake the commits please? 

So probs
* Add TimeoutRetryOptions, MigrationTimeoutOptions, and _Parser
* Add MigrateRetryStrategy
* Implement migration logic

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: Good shout db207cdd0ac4b69b1ae4b9570f06191992fd74f5

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: yikes yes, too many stashes. Fixup: d33f59d847b2d8b0f92b886ed1badf7b000f2f80

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: I'll have to rewrite the first commits to sort of make a clean history. Let me have a go at this and see how it looks. I'll come back to you later.

delfick: perfect :)

marcelofern: @delfick this is done now (-:

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
```diff
@@ -67,6 +188,55 @@ def validate(self) -> None:
...
+        self.lock_retry_options.validate()
+
+
+class MigrateRetryStrategy:
+    timeout_options: MigrationTimeoutOptions
+    retries: int
+
+    def __init__(self, timeout_options: MigrationTimeoutOptions):
+        self.timeout_options = timeout_options
+        self.retries = 0
```
delfick, line 200: why not an dataclass.dataclass(frozen=True)?

tests/django_pg_migration_tools/management/commands/test_migrate_with_timeouts.py
```diff
@@ -70,17 +70,175 @@ def test_interface(self):
...
+    def test_retry_callback_is_called(self, mock_handle, mock_sleep):
+        mock_handle.side_effect = [timeouts.DBLockTimeoutError("Bang!")]
+        with pytest.raises(
+            Exception, match="Prove callback is called by blowing it up."
+        ):
+            management.call_command(
+                "migrate_with_timeouts",
+                lock_timeout_in_ms=50_000,
+                lock_timeout_max_retries=2,
+                retry_callback_path="tests.django_pg_migration_tools.management.commands.test_migrate_with_timeouts.example_callback",
```
delfick, line 166:
```suggestion
                retry_callback_path=f"{__name__}.example_callback",
```

marcelofern: Nice

fixup: 9c0784b98fb072ac5d419e86ddc7b60c31faf0cb

src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
marcelofern: I think there's something strange about putting that much logic behind a dataclass

PR #21, opened 2024-09-04
Developer experience improvements - RemoveIndexConcurrently hinting and docstring

s-cooper18:
When using the SaferRemoveIndexConcurrently, there was no docstring or any hints for what arguments should go into the class instantiation. This PR allows more helpful information to appear in my IDE such that it matches the hints the types of hints that appear for SaferAddIndexConcurrently.

What appears in VS Code

Before:
image

After:
image

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -12,7 +12,7 @@ babel==2.16.0
     # via sphinx
 build==1.2.1
     # via django-pg-migration-tools (pyproject.toml)
-certifi==2024.7.4
+certifi==2024.8.30
     # via requests
 cfgv==3.4.0
     # via pre-commit
@@ -42,6 +42,8 @@ docutils==0.20.1
     #   sphinx-rtd-theme
 environs==11.0.0
     # via django-pg-migration-tools (pyproject.toml)
+exceptiongroup==1.2.2
+    # via pytest
 filelock==3.15.4
     # via virtualenv
 identify==2.6.0
@@ -95,7 +97,7 @@ pytest==8.3.2
     # via
     #   django-pg-migration-tools (pyproject.toml)
     #   pytest-django
-pytest-django==4.8.0
+pytest-django==4.9.0
     # via django-pg-migration-tools (pyproject.toml)
 python-dotenv==1.0.1
     # via environs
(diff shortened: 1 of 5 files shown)


requirements/development.txt
```diff
@@ -134,10 +136,19 @@ sphinxcontrib-serializinghtml==2.0.0
...
 sqlparse==0.5.1
     # via django
+tomli==2.0.1
+    # via
+    #   build
+    #   django-stubs
+    #   mypy
+    #   nox
+    #   pytest
+    #   sphinx
```
marcelofern, line 146: While some of the upgrades on this file look alright, these new libraries seemed a bit suspect, so I went looking.

`tomli` is only required for Python < 3.11 see [[1](https://github.com/wntrblm/nox/blob/main/pyproject.toml#L45)][[2](https://github.com/sphinx-doc/sphinx/blob/master/pyproject.toml#L73)][[3](https://github.com/pytest-dev/pytest/blob/main/pyproject.toml#L55)]

This package is not required any more in recent versions of Python as [PEP 680](https://peps.python.org/pep-0680/) has added tomli to the standard library on Python 3.11.

I think it is possible that you are using an older version of Python and might need to update your Python binary.
That should be the reason why these updates are necessary on your machine.

For the purposes of these files in the master branch, I think we should always keep them aligned with the latest versions of Python we support, in this case Python 3.12.

src/django_pg_migration_tools/__init__.py
```diff
@@ -1,3 +1,3 @@
 """
-TODO Add a description of the package here.
+Tools to make Django migrations safer and more scalable.
```
marcelofern, line 2: Thank you 💖

src/django_pg_migration_tools/operations.py
```diff
@@ -192,6 +192,9 @@ class SaferRemoveIndexConcurrently(
     model_name: str
     name: str
 
+    def __init__(self, model_name: str, name: str) -> None:
+        return super().__init__(model_name=model_name, name=name)
+
```
marcelofern, line 197: That's an interesting one. Which IDE are you using?

It may be a bug on the IDE - or it may be caching files locally. There's no significant difference between `SaferRemoveIndexConcurrently` and `SaferAddIndexConcurrently` innit methods.

Keen to see if there's a fix for your IDE instead, so that we don't need to maintain code for a specific IDE in the repo.

marcelofern (commented):
Thanks for having a look into this, I've left a couple of comments below:

src/django_pg_migration_tools/operations.py
s-cooper18: I'm using Visual Studio Code - will investigate. I'll make a new PR with changes separated so they can be merged easily if they are good to go.

PR #24, opened 2024-09-17
Use --skip-empty for coverage report

marcelofern:
This is a better alternative to not running coverage against __init__.py files entirely. This could've been a problem if those files had code in it.

### DIFF
--- a/makefile
+++ b/makefile
@@ -29,14 +29,13 @@ test:
 .PHONY:coverage_run
 coverage_run:
 	coverage run --branch \
-	--omit="__init__.py" \
 	--include=src/django_pg_migration_tools/* \
 	--data-file=build/coverage \
 	-m pytest
 
 .PHONY:coverage_report
 coverage_report:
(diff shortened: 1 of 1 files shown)


DHUKK (approved):
🚀

delfick (approved):

meshy (approved):

meshy (commented):
Have you considered `--skip-covered` instead, so that this shows only files which need work?

marcelofern:
> Have you considered --skip-covered instead, so that this shows only files which need work?

Thanks for the suggestion, I will have a look with more detail.
The output looks like this:

```
coverage report --skip-covered --skip-empty --format= --data-file=build/coverage

Name                                        Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------------
src/django_pg_migration_tools/timeouts.py      73      5     36      2    92%
-----------------------------------------------------------------------------
TOTAL                                         165      5     52      2    96%

5 files skipped due to complete coverage.
```

PR #26, opened 2024-09-24
Safer unique constraints

marcelofern:
# Goal

The main goal of this PR is to add an operation `SaferAddUniqueConstraint` that allows users to create a unique constraint on an existing table without downtime. The justification goes as follow:

## Justification

The default operation used by Django (AddConstraint) performs a naive:

```sql
ALTER TABLE table ADD CONSTRAINT constraint UNIQUE (field);
```

This has several downsides:

- Blocks reads and writes.
- Can be blocked by other reads and writes and create a chain of locks.
- Doesn't work with retries (not idempotent).

This new class (`SaferAddUniqueConstraint`) provides a way to create a
unique constraint without locking reads/writes at all, by first creating
an UNIQUE index CONCURRENTLY, and using this index to generate the
constraint (which takes virtually no time) behind a lock-timeout.

Example usage in a migration:

```diff
  + from django_pg_migration_tools import operations
  from django.db import migrations, models

  class Migration(migrations.Migration):
  +   atomic = False

      dependencies = [("myapp", "0042_dependency")]

      operations = [
  -        migrations.AddConstraint(
  +        operations.SaferAddUniqueConstraint(
              model_name="mymodel",
              constraint=models.UniqueConstraint(fields=["foo"], name="foo_unique"),
          ),
      ]
```

## DDL queries

This new operation will trigger the following queries:

```sql
-- Check if the constraint already exists.
SELECT conname
FROM pg_catalog.pg_constraint
WHERE conname = 'foo_unique';

-- Necessary so that we can reset the value later.
SHOW lock_timeout;

-- Necessary to avoid lock timeouts. This is a safe operation as
-- CREATE UNIQUE INDEX CONCURRENTLY takes a weaker SHARE UPDATE EXCLUSIVE
-- lock.
SET lock_timeout = 0;

-- Check if an INVALID index already exists.
SELECT relname
FROM pg_class, pg_index
WHERE (
    pg_index.indisvalid = false
    AND pg_index.indexrelid = pg_class.oid
    AND relname = 'foo_unique_idx'
);

-- Remove the invalid index (only if the previous query returned one).
DROP INDEX CONCURRENTLY IF EXISTS foo_unique_idx;

-- Finally create the UNIQUE index
CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS foo_unique_idx ON myapp_mymodel;

-- Reset lock_timeout to its original value ("1s" as an example).
SET lock_timeout = '1s';

-- Start a transaction for the ALTER TABLE so that it can time out if
-- a long running transaction is active against the table.
BEGIN;

-- Perform the ALTER TABLE using the unique index just created.
ALTER TABLE "myapp_mymodel" ADD CONSTRAINT "foo_unique" UNIQUE USING INDEX "foo_unique_idx";

-- Finish up the transaction
COMMIT;
```

## Time-to-run comparison between ALTER TABLE, and UNIQUE INDEX + ALTER TABLE

image
image

## Outro

For more detailed information you can check [this blog post.](https://marcelofern.com/posts/postgres/unique-constraints-without-downtime/index.html)

### DIFF
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -6,6 +6,7 @@ from django.contrib.postgres import operations as psql_operations
 from django.db import migrations, models
 from django.db.backends.base import schema as base_schema
 from django.db.migrations.operations import base as base_operations
+from django.db.migrations.operations import models as operation_models
 
 
 class BaseIndexOperation(
@@ -35,6 +36,7 @@ class BaseIndexOperation(
         from_state: migrations.state.ProjectState,
         to_state: migrations.state.ProjectState,
         index: models.Index,
+        unique: bool,
         model: type[models.Model],
     ) -> None:
         self._ensure_not_in_transaction(schema_editor)
@@ -50,6 +52,9 @@ class BaseIndexOperation(
         index_sql = index_sql.replace(
             "CREATE INDEX CONCURRENTLY", "CREATE INDEX CONCURRENTLY IF NOT EXISTS"
         )
+        if unique:
+            index_sql = index_sql.replace("CREATE INDEX", "CREATE UNIQUE INDEX")
+
         schema_editor.execute(index_sql)
         self._ensure_original_lock_timeout_is_reset(schema_editor)
 
@@ -157,6 +162,7 @@ class SaferAddIndexConcurrently(
             to_state=to_state,
             index=self.index,
             model=model,
+            unique=False,
         )
 
     def database_backwards(
@@ -226,5 +232,171 @@ class SaferRemoveIndexConcurrently(
             to_state=to_state,
             index=index,
             model=model,
-            operation=self,
+            unique=False,
+        )
+
+
+class ConstraintOperationError(Exception):
+    pass
+
+
+class ConstraintAlreadyExists(ConstraintOperationError):
+    pass
+
+
+class SaferAddUniqueConstraint(BaseIndexOperation, operation_models.AddConstraint):
+    model_name: str
+    constraint: models.UniqueConstraint
+    raise_if_exists: bool
+
+    _CHECK_EXISTING_CONSTRAINT_QUERY = dedent("""
+        SELECT conname
+        FROM pg_catalog.pg_constraint
+        WHERE conname = %(constraint_name)s;
+    """)
+
+    def __init__(
+        self,
+        model_name: str,
+        constraint: models.UniqueConstraint,
+        raise_if_exists: bool = True,
+    ) -> None:
+        self.raise_if_exists = raise_if_exists
+        self.constraint = constraint
+
+        # Perform a basic input-validation at initialisation time rather than
+        # at migration time ("database_forwards"/"database_backwards") so that
+        # all the errors we __can__ raise at initialisation time are raised
+        # when the programmer is still coding rather than when the migration
+        # effectively runs. This is the "Django way" of doing it, as other
+        # operation classes take this same approach. See "AddIndex" operation.
+        self._validate()
+
+        super().__init__(model_name, constraint)
+
+    def database_forwards(
+        self,
+        app_label: str,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+        from_state: migrations.state.ProjectState,
+        to_state: migrations.state.ProjectState,
+    ) -> None:
+        if not self._can_migrate(app_label, schema_editor, from_state):
+            return
+
+        if not self._can_create_constraint(schema_editor):
+            return
+
+        index = self._get_index_for_constraint()
+        model = to_state.apps.get_model(app_label, self.model_name)
+        self.safer_create_index(
+            app_label=app_label,
+            schema_editor=schema_editor,
+            from_state=from_state,
+            to_state=to_state,
+            index=index,
+            model=model,
+            unique=True,
+        )
+
+        # Django doesn't have a handy flag "using=..." so we need to alter the
+        # SQL statement manually. We go from a SQL that looks like this:
+        #
+        # - ALTER TABLE "table" ADD CONSTRAINT "constraint" UNIQUE ("field")
+        #
+        # Into a SQL that looks like:
+        #
+        # - ALTER TABLE "table" ADD CONSTRAINT "constraint" UNIQUE USING INDEX "idx"
+        base_sql = str(self.constraint.create_sql(model, schema_editor))
+        alter_table_sql = base_sql.split(" UNIQUE")[0]
+        sql = f'{alter_table_sql} UNIQUE USING INDEX "{index.name}"'
+
+        # Now we can execute the schema change. We have lock timeouts back in
+        # place after creating the index that would prevent this operation from
+        # running for too long if it's blocked by another query. Otherwise,
+        # this operation should actually be quite fast - if it's not blocked -
+        # since we have created the unique index in the previous step.
+        return schema_editor.execute(sql)
+
+    def database_backwards(
+        self,
+        app_label: str,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+        from_state: migrations.state.ProjectState,
+        to_state: migrations.state.ProjectState,
+    ) -> None:
+        if not self._can_migrate(app_label, schema_editor, from_state):
+            return
+
+        if not self._constraint_exists(schema_editor):
+            # Nothing to delete.
+            return
+
+        # Execute the DDL that will drop the constraint.
+        super().database_backwards(
+            app_label=app_label,
+            schema_editor=schema_editor,
+            from_state=from_state,
+            to_state=to_state,
+        )
+
+    def _validate(self) -> None:
+        if not isinstance(self.constraint, models.UniqueConstraint):
+            raise ValueError(
+                "SaferAddUniqueConstraint only supports the UniqueConstraint class"
+            )
+
+    def _constraint_exists(
+        self,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+    ) -> bool:
+        cursor = schema_editor.connection.cursor()
+        cursor.execute(
+            self._CHECK_EXISTING_CONSTRAINT_QUERY,
+            {"constraint_name": self.constraint.name},
+        )
+        return bool(cursor.fetchone())
+
+    def _can_create_constraint(
+        self,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+    ) -> bool:
+        constraint_exists = self._constraint_exists(schema_editor)
+        if self.raise_if_exists and constraint_exists:
+            raise ConstraintAlreadyExists(
+                f"Cannot create a constraint with the name "
+                f"{self.constraint.name} because a constraint of the same "
+                f"name already exists. If you want to skip this operation "
+                f"when the constraint already exists, run the operation "
+                f"with the flag `skip_if_exists=True`."
+            )
+        # We can't re-create a constraint that already exists because the
+        # ALTER TABLE ... ADD CONSTRAINT is not idempotent.
+        return not constraint_exists
+
+    def _get_index_for_constraint(self) -> models.Index:
+        return models.Index(
+            *self.constraint.expressions,
+            fields=self.constraint.fields,
+            name=self.constraint.name,
+            condition=self.constraint.condition,
+            opclasses=self.constraint.opclasses,  # type: ignore[attr-defined]
+        )
+
+    def _can_migrate(
+        self,
+        app_label: str,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+        from_state: migrations.state.ProjectState,
+    ) -> bool:
+        model = from_state.apps.get_model(app_label, self.model_name)
+        return bool(self.allow_migrate_model(schema_editor.connection.alias, model))
+
+    def describe(self) -> str:
+        return (
+            f"Concurrently adds a UNIQUE index {self.constraint.name} on model "
+            f"{self.model_name} on field(s) {self.constraint.fields} if the "
+            f"index does not exist. Then, adds the constraint using the just-created "
+            f"index. NOTE: Using django_pg_migration_tools SaferAddUniqueConstraint "
+            f"operation."
         )
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -5,9 +5,13 @@ from textwrap import dedent
 from django.contrib.postgres import operations as psql_operations
 from django.db import migrations, models
 from django.db.backends.base import schema as base_schema
+from django.db.migrations.operations import base as base_operations
 
 
-class BaseIndexOperation:
+class BaseIndexOperation(
+    psql_operations.NotInTransactionMixin,
+    base_operations.Operation,
+):
     SHOW_LOCK_TIMEOUT_QUERY = "SHOW lock_timeout;"
 
     SET_LOCK_TIMEOUT_QUERY = "SET lock_timeout = %(lock_timeout)s;"
@@ -32,11 +36,10 @@ class BaseIndexOperation:
         to_state: migrations.state.ProjectState,
         index: models.Index,
         model: type[models.Model],
-        operation: SaferAddIndexConcurrently | SaferRemoveIndexConcurrently,
     ) -> None:
-        operation._ensure_not_in_transaction(schema_editor)
+        self._ensure_not_in_transaction(schema_editor)
 
-        if not operation.allow_migrate_model(schema_editor.connection.alias, model):
+        if not self.allow_migrate_model(schema_editor.connection.alias, model):
             return
 
         self._ensure_no_lock_timeout_set(schema_editor)
@@ -58,11 +61,10 @@ class BaseIndexOperation:
         to_state: migrations.state.ProjectState,
         index: models.Index,
         model: type[models.Model],
-        operation: SaferAddIndexConcurrently | SaferRemoveIndexConcurrently,
     ) -> None:
-        operation._ensure_not_in_transaction(schema_editor)
+        self._ensure_not_in_transaction(schema_editor)
 
-        if not operation.allow_migrate_model(schema_editor.connection.alias, model):
+        if not self.allow_migrate_model(schema_editor.connection.alias, model):
             return
 
         self._ensure_no_lock_timeout_set(schema_editor)
@@ -155,7 +157,6 @@ class SaferAddIndexConcurrently(
             to_state=to_state,
             index=self.index,
             model=model,
-            operation=self,
         )
 
     def database_backwards(
@@ -173,7 +174,6 @@ class SaferAddIndexConcurrently(
             to_state=to_state,
             index=self.index,
             model=model,
-            operation=self,
         )
 
 
@@ -207,7 +207,6 @@ class SaferRemoveIndexConcurrently(
             to_state=to_state,
             index=index,
             model=model,
-            operation=self,
         )
 
     def database_backwards(
--- a/tests/django_pg_migration_tools/test_operations.py
+++ b/tests/django_pg_migration_tools/test_operations.py
@@ -10,7 +10,7 @@ from django.db.migrations.state import (
     ModelState,
     ProjectState,
 )
-from django.db.models import Index
+from django.db.models import CheckConstraint, Index, Q, UniqueConstraint
 from django.test import override_settings, utils
 
 from django_pg_migration_tools import operations
@@ -35,6 +35,16 @@ WHERE (
 );
 """
 
+_CHECK_CONSTRAINT_EXISTS_QUERY = """
+SELECT conname
+FROM pg_catalog.pg_constraint cons
+JOIN pg_catalog.pg_class class ON class.oid = cons.conrelid
+WHERE (
+  class.relname = %(table_name)s
+  and conname = %(constraint_name)s
+);
+"""
+
 _CHECK_INVALID_INDEX_EXISTS_QUERY = """
 SELECT relname
 FROM pg_class, pg_index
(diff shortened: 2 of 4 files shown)


tests/django_pg_migration_tools/test_operations.py
```diff
@@ -427,3 +454,355 @@ def test_when_not_allowed_to_migrate(self):
...
+        )
+
+        # Proceed to add the unique index followed by the constraint:
+        with connection.schema_editor(atomic=False, collect_sql=False) as editor:
+            with utils.CaptureQueriesContext(connection) as queries:
+                operation.database_forwards(
+                    self.app_label, editor, project_state, new_state
+                )
+
+        # Assert the index has disappeared!!!
```
s-cooper18, line 551: ✨ Very useful comment here!

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -427,3 +454,355 @@ def test_when_not_allowed_to_migrate(self):
...
+        operation = operations.SaferAddUniqueConstraint(
+            raise_if_exists=False,
+            model_name="intmodel",
+            constraint=UniqueConstraint(
+                fields=("int_field",),
+                name="unique_int_field",
+            ),
+        )
+
+        with connection.schema_editor(atomic=False, collect_sql=False) as editor:
```
s-cooper18, line 764: ❓ So in this scenario this is our way to make it "idempotent", by causing the migration to "pass" even if the constraint already exists?

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -427,3 +454,355 @@ def test_when_not_allowed_to_migrate(self):
                 {"table_name": "example_app_charmodel", "index_name": "char_field_idx"},
             )
             assert cursor.fetchone()
+
```
s-cooper18, line 457: Should there be a test for the scenario where the index does not already exist initially for completeness' sake?

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+    ) -> None:
+        if not self._can_migrate(app_label, schema_editor, from_state):
+            return
+
+        if not self._can_create_constraint(schema_editor):
+            return
+
+        index = self._get_index_for_constraint()
+        model = to_state.apps.get_model(app_label, self.model_name)
+        self.safer_create_index(
```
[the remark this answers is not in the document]
marcelofern: Correct, the `safer_create_index` bit handles what it needs for it to be safe, meaning:

- setting the lock_timeout to 0.
- checking if an invalid index of the same name exists and removing it if so.
- creating the unique index concurrently.
- resetting the lock_timeout to the original value.

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+        app_label: str,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+        from_state: migrations.state.ProjectState,
+        to_state: migrations.state.ProjectState,
+    ) -> None:
+        if not self._can_migrate(app_label, schema_editor, from_state):
+            return
+
+        if not self._can_create_constraint(schema_editor):
+            return
```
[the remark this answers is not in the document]
marcelofern: Not quite, Django doesn't provide a framework for informing the caller back on what's happened. It only guarantees the model state reflects the database schema (with a few exceptions like indexes). In this case above the database schema and model state are in sync.

Some Django specific operations like index removal add the `if exists` conditional and do not provide feedback on what has been done. The best way to find this out is via `sqlmigrate`

tests/django_pg_migration_tools/test_operations.py
marcelofern: Correct.

As this migration can be retried or used as part of a larger migration file with multiple non-atomic operations (any in which could fail and require retries on their own ways), this is a way to make sure we don't blow up and require manual database intervention if the constraint already exists.

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+
+class ConstraintOperationError(Exception):
+    pass
+
+
+class ConstraintAlreadyExists(ConstraintOperationError):
+    pass
+
+
+class SaferAddUniqueConstraint(BaseIndexOperation, operation_models.AddConstraint):
```
delfick, line 251: This example code is very rough, but I feel it would be clearer what is part of the classes you're inheriting and what is part of the logic you're adding on top if that logic was isolated

<details>
<summary>suggestion</summary>

```
class _SaferAddUniqueConstraintLogic:
    model_name: str
    constraint: models.UniqueConstraint
    raise_if_exists: bool
    index_name: str

    CHECK_EXISTING_CONSTRAINT_QUERY: ClassVar[str] = dedent("""
        SELECT conname
        FROM pg_catalog.pg_constraint
        WHERE conname = %(constraint_name)s;
    """)

    def __init__(
        self,
        model_name: str,
        constraint: models.UniqueConstraint,
        raise_if_exists: bool = True,
    ) -> None:
        self.raise_if_exists = raise_if_exists
        self.constraint = constraint
        self._index_name = f"{constraint.name}_idx"

    def execute_forwards(
        self,
        *,
        model: object,
        schema_editor: base_schema.BaseDatabaseSchemaEditor,
        index: str,
    ) -> None:
        # Django doesn't have a handy flag "using=..." so we need to alter the
        # SQL statement manually. We go from a SQL that looks like this:
        #
        # - ALTER TABLE "table" ADD CONSTRAINT "constraint" UNIQUE ("field")
        #
        # Into a SQL that looks like:
        #
        # - ALTER TABLE "table" ADD CONSTRAINT "constraint" UNIQUE USING INDEX "idx"
        base_sql = str(self.constraint.create_sql(model, schema_editor))
        alter_table_sql = base_sql.split(" UNIQUE")[0]
        sql = f'{alter_table_sql} UNIQUE USING INDEX "{index.name}"'

        # Run the sql to add the constraint inside a transaction. We have lock
        # timeouts back in place after creating the index that would prevent
        # this operation from running for too long if it's blocked by another
        # query. Otherwise, this operation should actually be quite fast - if
        # it's not blocked - since we have created the unique index in the
        # previous step.
        with transaction.atomic():
            return schema_editor.execute(sql)

    def constraint_exists(
        self,
        schema_editor: base_schema.BaseDatabaseSchemaEditor,
    ) -> bool:
        cursor = schema_editor.connection.cursor()
        cursor.execute(
            self._CHECK_EXISTING_CONSTRAINT_QUERY,
            {"constraint_name": self.constraint.name},
        )
        return bool(cursor.fetchone())

    def can_create_constraint(
        self,
        schema_editor: base_schema.BaseDatabaseSchemaEditor,
    ) -> bool:
        constraint_exists = self._constraint_exists(schema_editor)
        if self.raise_if_exists and constraint_exists:
            raise ConstraintAlreadyExists(
                f"Cannot create a constraint with the name "
                f"{self.constraint.name} because a constraint of the same "
                f"name already exists. If you want to skip this operation "
                f"when the constraint already exists, run the operation "
                f"with the flag `skip_if_exists=True`."
            )
        # We can't re-create a constraint that already exists because the
        # ALTER TABLE ... ADD CONSTRAINT is not idempotent.
        return not constraint_exists

    def get_index_for_constraint(self) -> models.Index:
        return models.Index(
            *self.constraint.expressions,
            fields=self.constraint.fields,
            name=self._index_name,
            condition=self.constraint.condition,
            opclasses=self.constraint.opclasses,  # type: ignore[attr-defined]
        )

    def validate(self) -> None:
        if not isinstance(self.constraint, models.UniqueConstraint):
            raise ValueError(
                "SaferAddUniqueConstraint only supports the UniqueConstraint class"
            )

        if len(self._index_name) > MAX_RELATION_NAME_LEN:
            raise ValueError(
                f"Index name {self._index_name} is longer than "
                f"{MAX_RELATION_NAME_LEN} characters limit."
            )


class SaferAddUniqueConstraint(BaseIndexOperation, operation_models.AddConstraint):
    logic: _SaferAddUniqueConstraintLogic

    def __init__(
        self,
        model_name: str,
        constraint: models.UniqueConstraint,
        raise_if_exists: bool = True,
    ) -> None:
        self._logic = _SaferAddUniqueConstraintLogic(
            model_name=model_name,
            constraint=constraint,
            raise_if_exists=raise_if_exists,
        )

        # Perform a basic input-validation at initialisation time rather than
        # at migration time ("database_forwards"/"database_backwards") so that
        # all the errors we __can__ raise at initialisation time are raised
        # when the programmer is still coding rather than when the migration
        # effectively runs. This is the "Django way" of doing it, as other
        # operation classes take this same approach. See "AddIndex" operation.
        self._logic.validate()

        super().__init__(model_name, constraint)

    def database_forwards(
        self,
        app_label: str,
        schema_editor: base_schema.BaseDatabaseSchemaEditor,
        from_state: migrations.state.ProjectState,
    ) -> None:
        if not self._can_migrate(app_label, schema_editor, from_state):
            return

        if not self._logic.can_create_constraint(schema_editor):
            return

        index = self._logic.get_index_for_constraint()
        model = to_state.apps.get_model(app_label, self.model_name)
        self.safer_create_index(
            app_label=app_label,
            schema_editor=schema_editor,
            from_state=from_state,
            to_state=to_state,
            index=index,
            model=model,
            unique=True,
        )

        self._logic.execute_forwards(
            model=model, schema_editor=schema_editor, index=index
        )

    def database_backwards(
        self,
        app_label: str,
        schema_editor: base_schema.BaseDatabaseSchemaEditor,
        from_state: migrations.state.ProjectState,
    ) -> None:
        if not self._can_migrate(app_label, schema_editor, from_state):
            return

        if not self._logic.constraint_exists(schema_editor):
            # Nothing to delete.
            return

        # Drop the constraint in a transaction so that it can raise a lock
        # timeout error if it takes too long. This transaction.atomic ctx
        # manager is required because the "forward" part of the migration
        # requires an atomic=False flag.
        with transaction.atomic():
            super().database_backwards(
                app_label=app_label,
                schema_editor=schema_editor,
                from_state=from_state,
                to_state=to_state,
            )

    def _can_migrate(
        self,
        app_label: str,
        schema_editor: base_schema.BaseDatabaseSchemaEditor,
        from_state: migrations.state.ProjectState,
    ) -> bool:
        model = from_state.apps.get_model(app_label, self.model_name)
        return bool(self.allow_migrate_model(schema_editor.connection.alias, model))

    def describe(self) -> str:
        return (
            f"Concurrently adds a UNIQUE index {self._index_name} on model "
            f"{self.model_name} on field(s) {self.constraint.fields} if the "
            f"index does not exist. Then, adds the constraint using the just-created "
            f"index. NOTE: Using django_pg_migration_tools SaferAddUniqueConstraint "
            f"operation."
        )

```
</details>

tests/django_pg_migration_tools/test_operations.py
delfick, line 1: I love the comments on the tests

src/django_pg_migration_tools/operations.py
marcelofern: @delfick How about using composition over of inheritance? Making so that the `BaseIndexOperation` is the part being composed instead?


```python
class SaferAddIndexConcurrently(psql_operations.AddIndexConcurrently):
    def __init__(self, args, kwargs):
        self.index_operator = BaseIndexOperation()
        ....

class SaferRemoveIndexConcurrently(psql_operations.AddIndexConcurrently):
    def __init__(self, args, kwargs):
        self.index_operator = BaseIndexOperation()
        ....

class SaferAddUniqueConstraint(...):
    def __init__(self, args, kwargs):
        self.index_operator = BaseIndexOperation()
        ....
```

Alternatively, we can just turn those into class methods and use that class directly:

```python
class SaferRemoveIndexConcurrently(psql_operations.AddIndexConcurrently):
    def database_forwards(...):
        BaseIndexOperation.safer_create_index(...)
```

That achieves what you are suggesting by making it explicit when we are using "outside logic", and we get away from multiple inheritance.

delfick: yeah, exactly, the point is to have less things combined in one place.

marcelofern: Thanks, I've created this issue to track that work -> https://github.com/kraken-tech/django-pg-migration-tools/issues/28

delfick: ah yeah, with the intention of merging this and doing that kind of transformation on this class as separate PR after this one?

marcelofern: Yep 👍🏽 reminding that the refactor will touch more than this class as it needs to bring the other two operations inheriting from the BaseIndexOperation along. I also want to rename a few things as that base class doesn't feel like the right name so will do that in a follow up.

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -427,3 +454,355 @@ def test_when_not_allowed_to_migrate(self):
...
+        # Proceed to add the unique index followed by the constraint:
+        with connection.schema_editor(atomic=False, collect_sql=False) as editor:
+            with utils.CaptureQueriesContext(connection) as queries:
+                operation.database_forwards(
+                    self.app_label, editor, project_state, new_state
+                )
+
+        # Assert the index has disappeared!!!
+        # Be no alarmed, this seems a bit strange, but Postgres has "swapped"
```
delfick, line 552: should this be "Be not alarmed"?

delfick (approved):

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+        if not isinstance(self.constraint, models.UniqueConstraint):
+            raise ValueError(
+                "SaferAddUniqueConstraint only supports the UniqueConstraint class"
+            )
+
+        if len(self._index_name) > MAX_RELATION_NAME_LEN:
+            raise ValueError(
+                f"Index name {self._index_name} is longer than "
+                f"{MAX_RELATION_NAME_LEN} characters limit."
+            )
```
timb07, line 365: As noted above, we've computed `self._index_name` from the supplied constraint name; furthermore, the user (developer) has no insight into how we've done that.

This error message tells the user that the index name we've constructed is too long — but the real issue is that the supplied constraint name is longer than 59 characters, which is the arbitrary limit we've imposed by how we've calculated the index name.

I'm not sure the best way to fix this. Could we use the constraint name as the index name directly? Would that work? And would there be an issue? (Constraints and indexes live in separate namespaces in Postgres, so I think it should work.)

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -427,3 +454,355 @@ def test_when_not_allowed_to_migrate(self):
...
+        # the index for a constraint instead. Keeping both the index and the
+        # constraint would result in unnecessarily occupied space.
```
timb07, line 554: 🐼  I don't think unnecessary space usage is the issue. Rather, when the constraint is created using the index, the index becomes owned by the constraint, and therefore cannot be manipulated independently as an index anymore. The way to enforce that is to remove it from the catalog of indexes — i.e., it disappears.

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -427,3 +454,355 @@ def test_when_not_allowed_to_migrate(self):
...
+    def test_when_name_too_long(self):
+        project_state = ProjectState()
+        project_state.add_model(ModelState.from_model(IntModel))
+
+        with pytest.raises(ValueError):
+            operations.SaferAddUniqueConstraint(
+                model_name="intmodel",
+                constraint=UniqueConstraint(
+                    fields=("int_field",),
+                    name="a" * (operations.MAX_RELATION_NAME_LEN + 1),
```
timb07, line 806: In reality, as noted above, this should fail if the name is `operations.MAX_RELATION_NAME_LEN - 3` characters long.

timb07 (commented):
As noted, I don't see why we need to open a transaction for the ALTER TABLE; and the name length limit handling is a bit problematic.

tests/django_pg_migration_tools/test_operations.py
marcelofern: It should (-: fixup: 3ec5aea02bfe2b24c3a3f864e368eae518fdc219

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+        # ALTER TABLE ... ADD CONSTRAINT is not idempotent.
+        return not constraint_exists
+
+    def _get_index_for_constraint(self) -> models.Index:
+        return models.Index(
+            *self.constraint.expressions,
+            fields=self.constraint.fields,
+            name=self._index_name,
+            condition=self.constraint.condition,
+            opclasses=self.constraint.opclasses,  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
marcelofern: Some of these have disappeared with an update of mypy + django stubs. But it seems like some are still required.
I've checked that the interface is correct though, I think it's just a limitation of the underlying packages.

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+        sql = f'{alter_table_sql} UNIQUE USING INDEX "{index.name}"'
+
+        # Run the sql to add the constraint inside a transaction. We have lock
+        # timeouts back in place after creating the index that would prevent
+        # this operation from running for too long if it's blocked by another
+        # query. Otherwise, this operation should actually be quite fast - if
+        # it's not blocked - since we have created the unique index in the
+        # previous step.
+        with transaction.atomic():
+            return schema_editor.execute(sql)
```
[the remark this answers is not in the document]
marcelofern: Agreed, got stuck in the wrong track. Plus, implicit transactions are opened for individual queries so this is not necessary.

> PostgreSQL actually treats every SQL statement as being executed within a transaction. If you do not issue a BEGIN command, then each individual statement has an implicit BEGIN and (if successful) COMMIT wrapped around it. A group of statements surrounded by BEGIN and COMMIT is sometimes called a transaction block. [[1](https://www.postgresql.org/docs/17/tutorial-transactions.html)]

Fixups:
- 79bc8bd96e4e85d49670d8c5523f36345097c8fe
- d758a894444a847ddc64c30d937cb2ab8510ec79

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+
+        if not self._constraint_exists(schema_editor):
+            # Nothing to delete.
+            return
+
+        # Drop the constraint in a transaction so that it can raise a lock
+        # timeout error if it takes too long. This transaction.atomic ctx
+        # manager is required because the "forward" part of the migration
+        # requires an atomic=False flag.
+        with transaction.atomic():
```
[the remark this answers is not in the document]
marcelofern: Addressed in the above comment: https://github.com/kraken-tech/django-pg-migration-tools/pull/26#discussion_r1790906398

src/django_pg_migration_tools/operations.py
marcelofern: Fixup: b3c627119240240411b7382bcf65c5c8a8896cdc

I had an issue when I tried reusing the index name. Not sure why I would go through such trouble otherwise. My shell output history says:

> psycopg.errors.DuplicateTable: relation "unique_int_field" already exists

I think it's just one of those cases of "tried this twice then got busy with something else before checking again".
Same name works:

```sql
-- test index name can be reused. Works on all version of Postgres supported.

CREATE TABLE example_table (
    id SERIAL PRIMARY KEY,
    int_field INT
);

CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS
unique_int_field ON example_table (int_field);

ALTER TABLE example_table
ADD CONSTRAINT unique_int_field UNIQUE USING INDEX unique_int_field;
```

tests/django_pg_migration_tools/test_operations.py
marcelofern: I've removed this comment as part of https://github.com/kraken-tech/django-pg-migration-tools/commit/b3c627119240240411b7382bcf65c5c8a8896cdc

tests/django_pg_migration_tools/test_operations.py
marcelofern: Addressed via: https://github.com/kraken-tech/django-pg-migration-tools/pull/26#discussion_r1790934806

tests/django_pg_migration_tools/test_operations.py
marcelofern: Added here: 6298fed4e45251df2163a7499c38b0fc0920f854

src/django_pg_migration_tools/operations.py
```diff
@@ -226,5 +236,185 @@ def database_backwards(
...
+
+    def __init__(
+        self,
+        model_name: str,
+        constraint: models.UniqueConstraint,
+        raise_if_exists: bool = True,
+    ) -> None:
+        self.raise_if_exists = raise_if_exists
+        self.constraint = constraint
+        self._index_name = f"{constraint.name}_idx"
```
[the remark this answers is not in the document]
marcelofern: Addressed in: https://github.com/kraken-tech/django-pg-migration-tools/pull/26#discussion_r1790934806

src/django_pg_migration_tools/operations.py
timb07: With the fixup, `MAX_RELATION_NAME_LEN` is now no longer used and can be removed.

src/django_pg_migration_tools/operations.py
```diff
@@ -5,9 +5,18 @@
 from django.contrib.postgres import operations as psql_operations
 from django.db import migrations, models
 from django.db.backends.base import schema as base_schema
+from django.db.migrations.operations import base as base_operations
+from django.db.migrations.operations import models as operation_models
 
 
-class BaseIndexOperation:
+# Postgres doesn't allow any relation names that are longer than 63 bytes.
+MAX_RELATION_NAME_LEN = 63
```
timb07, line 13: This constant is now no longer used, and can be omitted.

timb07 (approved):
The revisions all look good. There's one further quick fixup left, but it should be good to go after that.

src/django_pg_migration_tools/operations.py
marcelofern: Good catch: 3621d43efd65e21f65f4c14de4f3a17d7cc948fc

PR #29, opened 2024-10-04
Improve coverage

marcelofern:
This will improve the coverage of the `apply_timeouts`, which is the only module we have with less than 100% coverage.

These changes add three scenarios that were acused by the coverage report of not being run:
    
    1. A happy path when no timeout exceptions are raised.
    2. When the inner code raises a transaction leak and we are allowed to
       close the leaky transaction.
    3. When the inner code raises a transaction leak and we are **NOT**
       allowed to close the leaky transaction.

### DIFF
--- a/tests/django_pg_migration_tools/test_timeouts.py
+++ b/tests/django_pg_migration_tools/test_timeouts.py
@@ -2,9 +2,47 @@ import datetime
 from unittest import mock
 
 import pytest
-from django.db import utils
+from django.db import connection, connections, transaction, utils
+from django.test import utils as test_utils
 
 from django_pg_migration_tools import timeouts
+
+
+class _Leaky:
+    """
+    A context manager that leaks a transaction when it does not exit
+    successfully.
+
+    This simulates a corner-case scenario where a leaky transaction falls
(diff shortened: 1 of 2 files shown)


timb07 (approved):
Ah, these tricky leaky transactions again…

LGTM (mostly on the basis of the tests passing)

tests/django_pg_migration_tools/test_timeouts.py
```diff
@@ -2,11 +2,49 @@
...
 
 import pytest
-from django.db import utils
+from django.db import connection, connections, transaction, utils
+from django.test import utils as test_utils
 
 from django_pg_migration_tools import timeouts
 
 
+class _Leaky:
```
delfick, line 11: might as well do

```
@contextlib.contextmanager
def _leaky():
    with transaction.atomic("default"):
        with connection.cursor() as cursor:
          # This will raise an exception that aborts the transaction
          # because the IntModel model has `int_field` as required field
          # and we aren't giving it.
          cursor.execute("INSERT INTO example_app_intmodel (id) VALUES (%s)", [1])

PR #30, opened 2024-10-07
Use specific exc for too many retries

marcelofern:
When we reach the maximum amount of retries we want to raise an specific exception type so that it doesn't get mixed with other CommandError's.


timb07 (approved):
LGTM

PR #32, opened 2024-10-08
Expose migration stdout to RetryState

marcelofern:
The `migrate_with_timeouts` command can yield control to a callback every time a retry attempt happens.

Prior to this commit, the argument passed to the callback didn't include the stdout.

This prevented callbacks of handling the rich output that Django provides.

This particularly means that retry_callbacks can't understand which migration failed and is now being retried. In turn, the callback can't log or do anything about it.

With these changes, we pass the `io.StringIO` object that is an argument into the migrate command so that callbacks can use it as they wish.

The stdout will include an output like this:

    Operations to perform:
      Apply all migrations: ...
    Running migrations:
      Applying foo.0042_foo... OK
      Applying bar.0777_bar...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


timb07 (approved):
LGTM

Issue #37, opened 2024-10-14
Make command to install the package in editable mode


marcelofern:
This will allow for a more iterative approach when performing the developer loop of:

- change the code
- test it
- repeat

For more information about development mode please visit:
- https://setuptools.pypa.io/en/latest/userguide/development_mode.html

marcelofern:
Resolved in https://github.com/kraken-tech/django-pg-migration-tools/pull/38

PR #38, opened 2024-10-14
Install package in editable mode

marcelofern:
For new contributors installing their development environment via `make dev`, this will allow them to perform changes on the repo without reinstalling the package via `pip install .`.

This allows exploration, experimentation, and iterative changes to take place without having to repeat the pip install command too many times.

From the `pip install` man page:

```
-e, --editable <path/url>
      Install a project in editable mode (i.e. setuptools "develop
      mode") from a local project path or a VCS url.

      (environment variable: PIP_EDITABLE)
```

This change addresses: https://github.com/kraken-tech/django-pg-migration-tools/issues/37


delfick (approved):
if we switch to `uv sync` and uv project stuff, then it'll do that for you

Issue #28, opened 2024-10-04
Encapsulate logic from BaseIndexOperation


marcelofern:
# Description

At present, classes needing the behaviour of `safer_create_index` and `safer_drop_index` are **_inheriting_** from `BaseIndexOperation`.

To better separate what the behaviour of the class is vs what's inherited, we want to transform that inheritance into composition. For example:

```python
# Using composition
class SaferAddIndexConcurrently(psql_operations.AddIndexConcurrently):
    def __init__(self, args, kwargs):
        self.index_operator = BaseIndexOperation()
        ....

class SaferRemoveIndexConcurrently(psql_operations.AddIndexConcurrently):
    def __init__(self, args, kwargs):
        self.index_operator = BaseIndexOperation()
        ....

class SaferAddUniqueConstraint(...):
    def __init__(self, args, kwargs):
        self.index_operator = BaseIndexOperation()
        ....
````

Another alternative is to use the methods under `BaseIndexOperation` as class methods and run them directly:

```python
class SaferRemoveIndexConcurrently(psql_operations.AddIndexConcurrently):
    def database_forwards(...):
        BaseIndexOperation.safer_create_index(...)
```

PR #39, opened 2024-10-15
Encapsulate index creation logic

marcelofern:
Addresses https://github.com/kraken-tech/django-pg-migration-tools/issues/28

### DIFF
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -9,15 +9,13 @@ from django.db.migrations.operations import base as base_operations
 from django.db.migrations.operations import models as operation_models
 
 
-class SafeIndexOperationManager(
-    psql_operations.NotInTransactionMixin,
-    base_operations.Operation,
-):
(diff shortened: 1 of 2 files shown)


src/django_pg_migration_tools/operations.py
```diff
@@ -9,7 +9,7 @@
 from django.db.migrations.operations import models as operation_models
 
 
-class BaseIndexOperation(
+class SafeIndexOperationManager(
     psql_operations.NotInTransactionMixin,
     base_operations.Operation,
 ):
```
marcelofern: In a future refactoring I want to make this class stop inheriting from `NotInTransaction` and `Operation`.

We should have a `MigrationChecker` class that is responsible for addressing whether a migration can happen or not.
This will make it easier to share behaviour between safer index operations and safer constraint operations.

Most of the refactoring here will help with this move. In the interests of not having a big refactoring PR this will be addressed in -> https://github.com/kraken-tech/django-pg-migration-tools/issues/40

delfick (approved):
lovely :)

PR #41, opened 2024-10-21
Safer remove unique constraint

marcelofern:
## Description

The objective of this PR is to create a new operation called `SaferRemoveUniqueIndex` that provides the mirrored functionality of `SaferAddUniqueIndex`..

## Why?
One could use Django's default `RemoveConstraint` operation which would perform the following DDL directly:

```sql
ALTER TABLE "table" DROP CONSTRAINT "constraint";
```

And the reverse DDL is:

```sql
ALTER TABLE "table" ADD CONSTRAINT "constraint" UNIQUE ("field");
```

However...
  - Those operations are not idempotent.
  - The reverse operation in particular is extra unsafe. It will block writes and reads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- `SaferRemoveUniqueConstraint` operation was added. This is the complement for
+  `SaferAddUniqueConstraint` - but with the forward and backwards operations
+  swapped.
+
 ## [0.1.7] - 2024-10-09
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


delfick (approved):
love it :)

I think my one suggestion is `SafeConstraintOperationManager.create_constraint` and `SafeConstraintOperationManager.drop_constraint` instead of `create` and `drop`

s-cooper18:
Looks good from my end and also agree with @delfick's comment about create_constraint and drop_constraint - I got a confused a few times reading it, thinking that `create()` was about instantiating the class rather than creating the constraint.

marcelofern:
Thanks @delfick and @s-cooper18 I've changed the function name:
- e93b261f4f7b74daca852dc0d3648d99ed199fc8
- 7d3f0fdb8a6be066214c738910ea20578d75e3ee

PR #42, opened 2024-10-29
Handle scenario where unique constraint with condition is provided

s-cooper18:
Postgres does not support unique constraints with conditions, however, Django allows it to be possible to specify a unique constraint with a condition. In postgres, these are represented as unique indexes with conditions instead

Here, we modify the operations to add or remove constraints to align with django operations behaviour - adding or removing the unique index - however with "if exists / not exists" and concurrently added to these.

An example of what is currently generated in Django when using a `migrations.AddConstraint(`

`CREATE UNIQUE INDEX "" ON "" ("") WHERE "status" IN (''ACCEPTED', 'GENERATED', 'STARTED');`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed a bug where `SaferAddUniqueConstraint` and `SaferRemoveUniqueConstraint` would
+  accept unique constraints with conditions on them, but produce invalid SQL. Unique
+  constraints with conditions are now handled as partial unique indexes, as per the
+  equivalent Django `AddConstraint` and `RemoveConstraint` operations.
+
 ### Added
 
 - `SaferRemoveUniqueConstraint` operation was added. This is the complement for
--- a/tests/django_pg_migration_tools/test_operations.py
+++ b/tests/django_pg_migration_tools/test_operations.py
@@ -14,7 +14,7 @@ from django.db.models import CheckConstraint, Index, Q, UniqueConstraint
 from django.test import override_settings, utils
 
 from django_pg_migration_tools import operations
-from tests.example_app.models import CharModel, IntModel
+from tests.example_app.models import AnotherCharModel, CharModel, IntModel
 
 
 _CHECK_INDEX_EXISTS_QUERY = """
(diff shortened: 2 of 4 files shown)


SynStratos (approved):
Great and neat job!

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -924,6 +924,153 @@ def test_when_not_unique_constraint(self):
...
+        # 5. Finally create the index concurrently.
+        assert (
+            queries[4]["sql"]
+            == 'CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "unique_int_field" ON "example_app_intmodel" ("int_field") WHERE "int_field" >= 2'
+        )
+        # 6. Set the timeout back to what it was originally.
+        assert queries[5]["sql"] == "SET lock_timeout = '1s';"
+
+        assert len(queries) == 6
+        # there is not constraint
```
marcelofern, line 1020: Maybe you meant to add a check here for the constraint? This comment seems a bit isolated. I think we could ✂️ as this is more of an "index test" than a constraint one.

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -924,6 +924,153 @@ def test_when_not_unique_constraint(self):
...
+
+            cursor.execute(
+                _CHECK_CONSTRAINT_EXISTS_QUERY,
+                {
+                    "table_name": "example_app_intmodel",
+                    "constraint_name": "unique_int_field",
+                },
+            )
+            assert cursor.fetchone() is None
+
```
marcelofern, line 1073: Suggestion: [✂️] there was never a constraint here, only an index. So probably better to remove this assertion.

marcelofern (commented):
That looks pretty good, thanks @s-cooper18 !
I've left a few minor comments.

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -1070,6 +1217,148 @@ def test_operation(self):
...
+        operation.state_forwards(self.app_label, new_state)
+        assert (
+            len(
+                new_state.models[self.app_label, "anothercharmodel"].options[
+                    "constraints"
+                ]
+            )
+            == 0
+        )
```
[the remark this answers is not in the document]
s-cooper18: ❓ Did you mean just the lines highlighted here or all the tests checking lines too? (Assuming this for now)

s-cooper18:
Updated to reflect feedback https://github.com/kraken-tech/django-pg-migration-tools/pull/42/commits/7d7ce8c2579625d40a4546d30d7354b95c5e95c7

marcelofern:
@s-cooper18 thanks, that looks good to me. Can you also please append an entry in the CHANGELOG about this changes to this existing commit? Thanks.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
+## Fixed
```
marcelofern, line 11: Argh sorry @s-cooper18 this is missing an extra `#`

marcelofern (approved):

s-cooper18:
Would you be able to merge when you're ready please @marcelofern :) I don't have merge permissions

PR #45, opened 2024-11-04
Add SaferAlterFieldSetNotNull operation

marcelofern:
This commit adds a way to perform an ALTER TABLE ... NOT NULL safely.

In Django, when you set `null=False` on an existing column, Django will create the following migration operation:

```py
migrations.AlterField(
    model_name='foo',
    name='bar',
    # note that null=False is the default so it isn't
    # added in the operation
    field=models.IntegerField(),
)
```

This will trigger the following SQL:

```sql
ALTER TABLE "foo" ALTER COLUMN "bar_id" SET NOT NULL;
```

This is not a safe operation. It will perform a full table scan while holding an AccessExclusiveLock, blocking all reads and writes on the table.

The below is what this operation does instead:

```sql
-- The below still requires ACCESS EXCLUSIVE lock, but doesn't require a full
-- table scan.
-- This check will only be applied to new or modified rows, existing rows
-- won't be validated because of the NOT VALID clause.
ALTER TABLE foo
ADD CONSTRAINT bar_id_not_null
CHECK (bar_id IS NOT NULL) NOT VALID;

-- The below performs a sequential scan, but without an exclusive lock.
-- Concurrent sessions can read/write.
-- The operation will require a SHARE UPDATE EXCLUSIVE lock, which will block
-- only other schema changes and the VACUUM operation.
-- This may take a long time, but the operation is idempotent once the
--  constraint is marked as valid.
ALTER TABLE foo VALIDATE CONSTRAINT bar_id_not_null;

-- Requires ACCESS EXCLUSIVE LOCK, but bar_id_not_null proves that there is no
-- NULL in this column and a full table scan is not required. Therefore, the
-- ALTER TABLE command should be fast.
-- If you are not on Postgres >=12, you should skip this as it will take a lot
-- of time, and the current CHECK constraint might be good enough.
ALTER TABLE foo ALTER COLUMN bar_id SET NOT NULL;

-- The CHECK constraint has fulfilled its obligation and can now departure.
-- This takes an ACCESS EXCLUSIVE lock, but should run very fast.
ALTER TABLE foo DROP CONSTRAINT bar_id_not_null;
```


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- The `SaferAlterFieldSetNotNull` operation was added. This will safely change
+  a field from nullable to not nullable.
```
s-cooper18, line 14: 🐼 safely -> more safely - (nothing is 100% safe)

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+
+        # Just checking for NOT NULL isn't sufficient, we need to check that
+        # the constraint doesn't exist (VALID or NOT). Even a NOT VALID
+        # constraint would set the field to be non-nullable.
+        is_not_null = self._is_not_null(schema_editor, table_name, column_name)
+        constraint_exists = self._constraint_exists(schema_editor, constraint_name)
+        if is_not_null and (not constraint_exists):
+            return
+
+        if not constraint_exists:
```
s-cooper18, line 636: 🐼 If constraint is null and constraint doesn't exist

Is the flow more easily visible if it is like this?

```
if not constraint_exists:
    if is_not_null:
        return
        
   ...

```

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+        We need a unique name for the constraint.
+        We don't care too much about what the name itself turns out to be. This
+        constraint will be deleted at the end of the process anyway.
+        Here we just give it some resemblance to the table and column names the
+        constraint was created from.
+        """
+        suffix: str = django_backends_utils.names_digest(  # type:ignore[attr-defined]
+            table_name, column_name, length=10
+        )
+        return f"{table_name[:10]}_{column_name[:10]}_{suffix}"
```
s-cooper18, line 703: ❓ In the scenario that there is a not null constraint that already exists with a different name (i.e. was generated from something else), would the creation of a second null constraint (because this would only search for this specific constraint name) covering the same column cause any issues?

docs/usage/operations.rst
```diff
@@ -396,3 +396,112 @@ Class Definitions
                   name="foo_unique",
               ),
           ]
+
+
+.. py:class:: SaferAlterFieldSetNotNull(model_name: str, name: str)
+
+    Provides a way to safely alter a field to NOT NULL.
```
s-cooper18, line 403: ♻️ 🐼 "more safely"

s-cooper18 (commented):
Added a comment about an edge case

docs/usage/operations.rst
marcelofern: 5c1adbbca3422106d4be3ea378d8b64a6b06092c

src/django_pg_migration_tools/operations.py
marcelofern: That's a great question. Having an existing constraint wouldn't cause any issues here. You can have duplicate NOT NULL constraints on the same field as long as they have different names.

However, it is possible to re-use an existing constraint that a user created for whatever reason (maybe they even forgot about it).

But I don't think it should be our responsibility to clean up the schema for the user when the action was not caused by us (unless it blocks us in some way).

At the minimum, doing that would involve adding extra introspective queries to cover these corner-cases, thus penalising regular users that won't likely have an existing constraint against that field.

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+        else:
+            # Constraint exists and is NOT VALID.
+            self._validate_constraint(schema_editor, table_name, constraint_name)
+            self._alter_table_not_null(schema_editor, table_name, column_name)
+            self._alter_table_drop_constraint(
+                schema_editor, table_name, constraint_name
+            )
+            return
+
+    def _alter_table_not_valid_constraint(
```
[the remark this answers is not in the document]
marcelofern: a8e0d20b7ae34f35bec51f075fa1a28620fec1bc

src/django_pg_migration_tools/operations.py
marcelofern: I left these guards separated from each other because each one represents one distinct stage.

- Field is already NOT NULL, do nothing.
- Constraint doesn't exist yet.
- ... etc

But also my personal preference is to have flat code instead of nested `if` guards when possible

CHANGELOG.md
marcelofern: 922d2c84de74a3539d53460047253d026dd82c79

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+    ) -> None:
+        cursor = schema_editor.connection.cursor()
+        cursor.execute(
+            psycopg_sql.SQL(ConstraintQueries.ALTER_TABLE_VALIDATE_CONSTRAINT).format(
+                table_name=psycopg_sql.Identifier(table_name),
+                constraint_name=psycopg_sql.Identifier(constraint_name),
+            )
+        )
+
+    def set_null(
```
timb07, line 785: 🐼 It would be better for this "public" method to be at the top of the class, rather than after a number of "private" methods.

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+        table_name = model._meta.db_table
+        constraint_name = self._get_constraint_name(table_name, column_name)
+
+        # Just checking for NOT NULL isn't sufficient, we need to check that
+        # the constraint doesn't exist (VALID or NOT). Even a NOT VALID
+        # constraint would set the field to be non-nullable.
+        is_not_null = self._is_not_null(schema_editor, table_name, column_name)
+        constraint_exists = self._constraint_exists(schema_editor, constraint_name)
+        if is_not_null and (not constraint_exists):
+            return
```
timb07, line 634: Why are we doing these checks? (It's to enable reentrancy and idempotency.) The explanations here don't make much sense without that context.

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+class NullsManager(base_operations.Operation):
+    def set_not_null(
+        self,
+        app_label: str,
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+        model: type[models.Model],
+        column_name: str,
+    ) -> None:
+        """
+        The following routine takes into consideration:
```
timb07, line 597: What does this method do, and how?

The `if/elif` conditions in the code below (and the subseries of operations within each) are hard to understand without knowing what this method does, and the series of steps it takes to achieve that.

The above information is in the commit comment. It should be in this docstring instead. See [Wints' blogpost](https://codeinthehole.com/tips/explaining/#this-applies-everywhere).

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+            )
+            return
+        else:
+            # Constraint exists and is NOT VALID.
+            self._validate_constraint(schema_editor, table_name, constraint_name)
+            self._alter_table_not_null(schema_editor, table_name, column_name)
+            self._alter_table_drop_constraint(
+                schema_editor, table_name, constraint_name
+            )
+            return
```
timb07, line 659: At the expense of adding a `not` to the `elif` condition, I would swap these two cases. That way the three cases are in order of starting from scratch, picking up the process after 1 step was completed (adding the constraint), or after 2 steps were completed (adding the constraing and validating it).

⚠️  What if the process only completed 3 steps, resulting in the column being NOT NULL but still leaving the constraint behind? It looks like that case isn't dealt with.

docs/usage/operations.rst
```diff
@@ -396,3 +396,112 @@ Class Definitions
...
+      -- block only other schema changes and the VACUUM operation.
+      ALTER TABLE foo VALIDATE CONSTRAINT bar_not_null;
+
+      -- Requires ACCESS EXCLUSIVE LOCK, but bar_not_null proves that there
+      -- is no NULL in this column and a full table scan is not required.
+      -- Therefore, the ALTER TABLE command should be fast.
+      -- If you are not on Postgres <= 11, you should skip this as it will
+      -- take a lot of time, and the current CHECK constraint might be good
+      -- enough.
+      ALTER TABLE foo ALTER COLUMN bar SET NOT NULL;
```
timb07, line 452: Advising "you should skip this" doesn't make much sense, given that `SaferAlterFieldSetNotNull` doesn't provide a way to skip it.

timb07 (commented):
This is looking really good, but there are some things to address as noted.

src/django_pg_migration_tools/operations.py
```diff
@@ -525,3 +583,256 @@ def describe(self) -> str:
...
+        schema_editor: base_schema.BaseDatabaseSchemaEditor,
+        table_name: str,
+        column_name: str,
+    ) -> bool:
+        cursor = schema_editor.connection.cursor()
+        cursor.execute(
+            NullabilityQueries.IS_COLUMN_NOT_NULL,
+            {"table_name": table_name, "column_name": column_name},
+        )
+        return bool(cursor.fetchone())
```
timb07, line 690: Reviewer's [ 🍷 / 🧀 ]:

After executing that query which does a `SELECT 1 ... WHERE ...`, the `cursor.fetchone()` call returns either `None` or `(1,0)`. Even though those two possibilities are of different types, they correctly result in `False` or `True` when cast to `bool`.

(Reviewer is wary of this situation after creating a bug due to `bool((0,))` returning `True`.)

src/django_pg_migration_tools/operations.py
timb07: Thinking about this bit of code some more, I wonder whether it might be clearer (and also reduce code duplication) to structure it as follows:

```python
if need_to_do_step_1:
    do_step1()

if need_to_do_step_2:
    do_step2()

if need_to_do_step_3:
    do_step3()

if need_to_do_step_4:
    do_step4()
```

In normal operation, that would be sequence of four `if True` conditions, and all four steps would be done. But if there had been a failure previously, it would pick up with whichever step needed to be done next.

That might require some extra SQL queries though — I haven't checked.

src/django_pg_migration_tools/operations.py
```diff
@@ -34,6 +37,61 @@ class ConstraintQueries:
...
+        WHERE
+            conname = %(constraint_name)s
+            AND convalidated IS FALSE;
+    """)
+
+    ALTER_TABLE_CONSTRAINT_NOT_NULL_NOT_VALID = dedent("""
+        ALTER TABLE {table_name}
+        ADD CONSTRAINT {constraint_name}
+        CHECK ({column_name} IS NOT NULL) NOT VALID;
+    """)
```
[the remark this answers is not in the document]
marcelofern: > SQL string templates defined so far away from where they're used.

I see your point, and I agree (well, sort of). The same as above can be said
about Kraken's big int operations, and about Django's schema editor; this is an
observable pattern.

From previous reviews interactions, the code holding the template was
determined to not be responsible for executing the query itself. That
responsibility lies within each operation manager.

I can see the trade-off of both approaches, but architecturally I prefer the
current code. I know that all the SQL queries will only be executed from within
the Operation Managers and not in different places across the repo. I don't
know how I should address your reviewing concern. I think that the way I review
is different so I don't have the same issues.

src/django_pg_migration_tools/operations.py
marcelofern: My original intend was to explain that a NOT VALID constraint would set the
catalogue entry for this table attr to be NOT NULL, so querying for the
catalogue is not sufficient. I think this comment shouldn't be here anymore.
e2b4075df2c6de380a28426dd1875eb9e088018d

src/django_pg_migration_tools/operations.py
marcelofern: I'm concerned about adding how/what description of this routine in the
docstrings, I prefer context over how/what. Docstrings become out-of-date as a
function grows, inducing confusion and errors.

As of now the information is available in 3 places: commit message, docs, and
tests. I'd like to retain the information in the commit message because it
reflects the information at the time the change was made, and git blame is
very powerful for that.

src/django_pg_migration_tools/operations.py
marcelofern: >  leaving the constraint behind? It looks like that case isn't dealt with.

Good shout. That case is indeed missing.
8846dd2954b43cb86a27dfeaf5bbb8c620e8184a

> That might require some extra SQL queries though — I haven't checked.

They do. That's the reason I didn't want to do it in this way.

If there's a particular control flow that reduces the number of introspective
queries even further, I'm in. However I don't want to penalise the most common
case (where all steps need to be performed) with one extra introspection query
per step.

docs/usage/operations.rst
marcelofern: copy-pasta fixed in: b3f5693424df4f8f620308eb11c59058fce5911d

s-cooper18 (approved):
LGTM now

src/django_pg_migration_tools/operations.py
timb07: I think your argument is valid when applied to internal Kraken Tech repos, where the users of the code are all developers of the code, and will be accustomed to checking git history and commit messages.

But this is a third-party utility library. The users of this code will most likely have installed it via `pip install`; they won't clone the repo, and they won't look at the tests. They may view this method by following code definitions in their IDE. The docstring as it stands, without the context of even a single line statement of what it does, and without a reference to where to get more information, is unhelpful.

> Docstrings become out-of-date as a function grows, inducing confusion and errors.

There are ~18 lines of docstring below the one line I highlighted above, all of which are subject to your concern about becoming out of date.

I don't know if there's much overlap between a docstring you'd be happy with and one that would keep me happy also. :(

But as a thought experiment, imagine you'd released the code without the fix in https://github.com/kraken-tech/django-pg-migration-tools/commit/8846dd2954b43cb86a27dfeaf5bbb8c620e8184a. Then some user of this package encountered an error that left behind the constraint when applying the migration, which failed initially, and found that the constraint was still there after the migration applied successfully. They then went digging into the code to find what happened. What docstring would help them the most to find the issue?

src/django_pg_migration_tools/operations.py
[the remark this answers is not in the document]
timb07: I agree that this is a pattern that already exists elsewhere, so you can't be faulted for using the same pattern.

I would appreciate if you gave some more thought to the idea that there are two different types of entity here, but both are represented by strings. I.e., this point:

> As well as verifying the variable names used, I also need to verify the two types of substitution, i.e. parameter passing with %(something) via cursor.execute() and identifier substitution via {something} and SQL().format().

Is there some way that we could more clearly distinguish them?

In the end, any decent test would uncover any mistakes in the use of these strings containing SQL, so we're fairly safe here. But I feel there's still some scope for changes that will make code comprehension easier.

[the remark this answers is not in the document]
marcelofern: Sorry for not addressing the identifier question, I might not have understood
that it was separated from the code-proximity issue.

We know we can only use `%s` and `%(name)s` arguments on query parameters [[1](https://www.psycopg.org/psycopg3/docs/basic/params.html)].
That means things inside a "WHERE" or "VALUES" clause.

We can't use `%s` or `%(name)` to build the query statement itself (DDL or
not). We must use `{name}` with a SQL() and Identifier().

All the queries we have here that use `{}` are ALTER TABLE DDL statements.
This is clear from their names as they are prefixed by `ALTER_TABLE_`.

A neat feature of psycopg is that you always know you are passing either a
query value or a query constructor by the way the template is called.

We know we can't pass query parameters to an ALTER TABLE instruction, the query
must be built. You still have to check the identifiers match, but you won't get
away of this problem no matter what shape your code has.

Does this address your question? I am not sure what you mean by entities.
There will be template queries that only receive values (WHERE/VALUES), there
will be queries that only receive constructor arguments (AT), and queries that
perhaps receive both (though we don't use them here).

src/django_pg_migration_tools/operations.py
marcelofern: > There are ~18 lines of docstring below the one line I highlighted above, all of which are subject to your concern about becoming out of date.

Yeah, that's a logical flaw in my argument. Just having any docstring is enough
for it to become out of date. A small nuance though, is that by only having
context in the docstring (that re-entrancy and idempotency is expected), the
person who found the bug would know that the code didn't follow the expected
outcome. I assume that's how you come to that conclusion that it was indeed a
problem (as you are the real bug-finder yourself in that scenario)?

I'd be up for giving more context about expected outcome − for example that
queries are expected to not lock reads/writes (at least not for very long) −
but I don't think we should explain the SQL as detailed as the docs have. I'm
not sure if that's what you are suggesting?

I agree that users will inspect code, docs, commands, etc., in different ways.
But I don't think that replicating documentation to all different ways a user
may interact with the repo is the solution

src/django_pg_migration_tools/operations.py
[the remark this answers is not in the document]
timb07: > I am not sure what you mean by entities.
There will be template queries that only receive values (WHERE/VALUES), there
will be queries that only receive constructor arguments (AT), and queries that
perhaps receive both (though we don't use them here).

That's basically what I was getting at. Having the strings defined some distance from where they're used makes it harder to check how they should be used, but as you point out, all the strings that need to be used with `SQL(...).format(...)` are named `ALTER_TABLE_...`, which mostly addresses the issue.

src/django_pg_migration_tools/operations.py
marcelofern: Thanks. I'm more relieved now that it's clear this isn't about detailing out
information about the SQL queries, what they exactly do, and further context.

With a few small alterations for clarification here is the change: cbe227bcc8766f39404727b9a3460835b357a387.

timb07: Thanks, those clarifications LGTM.

timb07 (approved):
This looks good to me now. Thanks for going a few rounds of comments to reach this point. :)

PR #47, opened 2024-11-14
Pass in database name to RetryState

marcelofern:
Users with multi-db Django configurations may want to know exactly which
database failed to push the migration forward.

This PR adds a new field called "database" to the RetryState dataclass so
that users who are hooking up onto the retry callback can find that
information.


s-cooper18 (approved):
Makes sense

PR #48, opened 2024-11-18
Make it obvious that nox uses psycopg3

marcelofern:
TBD


src/django_pg_migration_tools/operations.py
```diff
@@ -9,7 +9,15 @@
 from django.db.migrations.operations import base as base_operations
 from django.db.migrations.operations import fields as operation_fields
 from django.db.migrations.operations import models as operation_models
-from psycopg import sql as psycopg_sql
+
+
+try:
+    from psycopg import sql as psycopg_sql
+except ImportError:  # pragma: no cover
```
marcelofern, line 16: [🍷] Author's notes:

While we now run tests against both psycopg2 and psycopg3, we only run the coverage job against psycopg3 to avoid complexity and costs of running another job. For now, this import is the only bit that has a difference between psycopg2 and psycopg3.

timb07 (approved):
I get the general gist of what's going on from the commit messages, and it makes sense to me.

The single change to production code looks fine to me, so I'm confident this PR doesn't break anything. :)

I think that's enough for a "LGTM". :p

PR #49, opened 2024-11-21
Prep for 0.1.10

marcelofern:


timb07 (approved):
Confirmed that deps are updated for both psycopg2 and psycopg3. 👍 

LGTM

PR #52, opened 2024-12-09
Safer check constraints

marcelofern:
# Safer Check Constraints

Django CHECK constraints can be created in the following way:

```python
class Meta:
    constraints = [
        models.CheckConstraint(
            check=Q(bar__gte=0),
            name='bar_not_negative',
        ),
    ]
```

In this case, Django will create the following migration operations:

```python
operations = [
    migrations.AddConstraint(
        model_name='foo',
        constraint=models.CheckConstraint(
            check=Q(bar__gte=0),
            name='bar_not_negative',
        ),
    ),
]
```

This operation will trigger an unsafe SQL query of the form:

```sql
ALTER TABLE foo
ADD CONSTRAINT bar_not_negative
CHECK (bar >= 0);
```

## Why is this unsafe

This operation acquires an AccessExclusiveLock, which is the most restricted
lock in Postgres, blocking any reads, writes, maintanence activities, and other
schema changes on the table.

It will also scan the whole table to make sure there are no violations in this
table. All that while holding onto that lock.

## A better way

First create a NOT VALID constraint. For example:

```sql
ALTER TABLE foo ADD CONSTRAINT bar_not_negative CHECK (bar >= 0) NOT VALID;
```

The constraint will only be enforced for new writes and won't have to scan
the whole table while holding the AccessExclusiveLock.

Then, you can validate the constraint in a separated step:

```sql
ALTER TABLE foo VALIDATE CONSTRAINT bar_not_negative;
```

This time Postgres will scan the whole table, but it will only acquire a
ShareUpdateExclusiveLock, which won't block any reads or writes on that table.

### DIFF
--- a/docs/usage/operations.rst
+++ b/docs/usage/operations.rst
@@ -621,3 +621,105 @@ Class Definitions
                   ),
               ),
           ]
+
+
+.. py:class:: SaferAddCheckConstraint(model_name: str, constraint: models.CheckConstraint)
+
+    Provides a safer way to add a check constraint to an existing model.
+
+    :param model_name: Model name in lowercase without underscores.
+    :type model_name: str
+    :param constraint: The object representing the constraint to add.
+    :type constraint: models.CheckConstraint
+
+    **Why use this SaferAddCheckConstraint operation?**
+    ---------------------------------------------------
+
+    When using Django's default ``AddConstraint`` operation, the SQL created
+    has the following form:
+
+    .. code-block:: sql
+
+      ALTER TABLE foo
+      ADD CONSTRAINT bar_not_negative
+      CHECK (bar >= 0);
+
+    This operation acquires an AccessExclusive lock, which is the most
+    constricted lock in Postgres, blocking any reads, writes, maintanence
+    activities, and other schema changes on the table.
+
+    It will also scan the whole table to make sure there are no violations of
+    the new constraint. All that while holding onto that lock.
+
+    This safer operation, will instead perform the following:
+
+    .. code-block:: sql
+
+      -- Add a NOT VALID constraint.
+      -- This type of constraint still works, but only for new writes.
+      -- It still requires the AccessExclusive lock, but as it doesn't need to
+      -- scan the table, it runs very fast.
+      ALTER TABLE foo
+      ADD CONSTRAINT bar_not_negative
+      CHECK (bar >= 0)
+      NOT VALID;
+
+      -- Validate the constraint.
+      -- This operation needs to scan the table, but it only holds a
+      -- ShareUpdateExclusive lock, which won't block reads or writes.
+      ALTER TABLE foo VALIDATE CONSTRAINT bar_not_negative;
+
+    Note: The operations above are not inside a transaction. This is by design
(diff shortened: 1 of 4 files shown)


tests/django_pg_migration_tools/test_operations.py
```diff
@@ -2701,3 +2702,347 @@ def test_operation_when_referred_model_is_defined_as_str(self):
...
+            'ALTER TABLE "example_app_intmodel" ADD CONSTRAINT "positive_int" '
+            'CHECK ("int_field" >= 0) NOT VALID;'
+        )
+
+        # 3. Validate it
+        assert queries[2]["sql"] == dedent("""
+            ALTER TABLE "example_app_intmodel"
+            VALIDATE CONSTRAINT "positive_int";
+        """)
+
```
timb07, line 2871: At this point it would be good to repeat the query from lines 2804-2809 along with an assertion that the constraint does now in fact exist. That would validate the creation of the constraint separately from the validation of the idempotency of the operation that's done next.

docs/usage/operations.rst
```diff
@@ -621,3 +621,100 @@ Class Definitions
...
+      CHECK (bar >= 0);
+
+    This operation acquires an AccessExclusive lock, which is the most
+    constricted lock in Postgres, blocking any reads, writes, maintanence
+    activities, and other schema changes on the table.
+
+    It will also scan the whole table to make sure there are no violations of
+    the new constraint. All that while holding onto that lock.
+
+    This safer operation, will insted perform the following:
```
timb07, line 654: typo: "instead"

docs/usage/operations.rst
```diff
@@ -621,3 +621,100 @@ Class Definitions
...
+      ALTER TABLE foo
+      ADD CONSTRAINT bar_not_negative
+      CHECK (bar >= 0)
+      NOT VALID;
+
+      -- Validate the constraint.
+      -- This operation needs to scan the table, but it only holds a
+      -- ShareUpdateExclusive lock, which won't block reads or writes.
+      ALTER TABLE foo VALIDATE CONSTRAINT bar_not_negative;
+
```
timb07, line 671: Is it worth adding something about why `atomic = False` (mentioned below) is necessary? That is, that to avoid holding the exclusive lock while the subsequent validation happens, they need to not be in a transaction together?

(My initial thought was that requiring `atomic = False` was unnecessary until I thought about this.)

timb07 (approved):
LGTM; I've made some minor suggestions.

tests/django_pg_migration_tools/test_operations.py
marcelofern: Fixup commit: 13c02ebdc1d2fa9b47674c3d494f8da5fb6b8e4c

docs/usage/operations.rst
marcelofern: Fixup commit: 2a1b8a0af3d47f13f51edb02989e5e90334e425f

docs/usage/operations.rst
marcelofern: Fixup commit: e7793abde9b20ae69c9cca64754af812cef2e94f

PR #55, opened 2024-12-10
Use schema_editor for executing DDLs

marcelofern:
# Fix strange behaviour with `sqlmigrate`

Django's schema_editor.execute command is special because when running sqlmigrate it only collects the SQL statements without running them.

This is different from getting a cursor and running the query against the database, which would immediately run the DDL no matter what.

This commit fixes so that all the DDLs are run through the schema_editor.

This commit does not include the instropection queries because that will be a more difficult fix. That is because the schema_editor.execute function does not return a cursor where we can call .fetchone or .fetchall. So if we can't fetch the result, we can answer the question from the instropection query.

There are way to override this behaviour with a few hacks, but since it is not immediately pressing this will have to come on a subsequent work unit.

### DIFF
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -26,7 +26,7 @@ except ImportError:  # pragma: no cover
 
 class TimeoutQueries:
     SHOW_LOCK_TIMEOUT = "SHOW lock_timeout;"
-    SET_LOCK_TIMEOUT = "SET lock_timeout = %(lock_timeout)s;"
+    SET_LOCK_TIMEOUT = "SET lock_timeout = {lock_timeout};"
 
 
 class IndexQueries:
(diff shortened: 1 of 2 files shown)


src/django_pg_migration_tools/operations.py
```diff
@@ -271,8 +271,11 @@ def safer_drop_index(
     def _set_lock_timeout(
         self, schema_editor: base_schema.BaseDatabaseSchemaEditor, value: str
     ) -> None:
-        cursor = schema_editor.connection.cursor()
-        cursor.execute(TimeoutQueries.SET_LOCK_TIMEOUT, {"lock_timeout": value})
+        schema_editor.execute(
+            psycopg_sql.SQL(TimeoutQueries.SET_LOCK_TIMEOUT)
+            .format(lock_timeout=psycopg_sql.Literal(value))
+            .as_string(schema_editor.connection.connection)
+        )
```
marcelofern, line 278: 🍷 `schema_editor.execute` takes a string. I had to transform that into a proper `SQL` composable object to avoid using f-strings with direct sql statements.

Issue #59, opened 2024-12-16
Add session-level advisory lock at the beginning of migrate_with_timeouts


marcelofern:
So that any race conditions can be avoided during migration

PR #61, opened 2024-12-17
Run migrate_with_timeouts with an advisory lock

marcelofern:
Prior to this change, the migrate_with_timeouts command could run in
parallel multiple times.

This is a problem because many migration operations that are multi-step
and non-atomic will behave badly under such circumstances.

To prevent this of happening, this command will now require a
session-level advisory lock for the entirety of its procedure.

The session-level advisory lock will be automatically removed once the
management command finishes, whether successfully or not.

See:
  https://github.com/django/django/blob/857b1048/django/core/management/base.py#L426

Issue:
  https://github.com/kraken-tech/django-pg-migration-tools/issues/59

## Demo

The shell at the top ran the command with the following migration

<details>
Migration:

```py
# Generated by Django 5.1.4 on 2024-12-17 20:03

from django.db import migrations


def sleep(*args, **kwargs):
    import time
    time.sleep(777)


class Migration(migrations.Migration):

    dependencies = [
        ('blog', '0001_initial'),
    ]

    operations = [migrations.RunPython(sleep)]

```
</details>

The shell at the top try to run the migration command again, and failed
immediatelly

image

### DIFF
--- a/tests/django_pg_migration_tools/management/commands/test_migrate_with_timeouts.py
+++ b/tests/django_pg_migration_tools/management/commands/test_migrate_with_timeouts.py
@@ -25,13 +25,14 @@ class TestMigrateWithTimeoutsCommand:
                 statement_timeout_in_ms=100_000,
             )
 
-        assert queries[0]["sql"] == "SHOW lock_timeout"
-        assert queries[1]["sql"] == "SET SESSION lock_timeout = '50000ms'"
-        assert queries[2]["sql"] == "SHOW statement_timeout"
-        assert queries[3]["sql"] == "SET SESSION statement_timeout = '100000ms'"
-        assert queries[4]["sql"] == "SET SESSION lock_timeout = '0'"
-        assert queries[5]["sql"] == "SET SESSION statement_timeout = '0'"
-        assert len(queries) == 6
+        assert queries[0]["sql"] == "SELECT pg_try_advisory_lock(8967004653001705610);"
+        assert queries[1]["sql"] == "SHOW lock_timeout"
+        assert queries[2]["sql"] == "SET SESSION lock_timeout = '50000ms'"
+        assert queries[3]["sql"] == "SHOW statement_timeout"
+        assert queries[4]["sql"] == "SET SESSION statement_timeout = '100000ms'"
+        assert queries[5]["sql"] == "SET SESSION lock_timeout = '0'"
+        assert queries[6]["sql"] == "SET SESSION statement_timeout = '0'"
+        assert len(queries) == 7
 
     @mock.patch("django.core.management.commands.migrate.Command.handle", autospec=True)
     @pytest.mark.django_db(transaction=True)
@@ -61,10 +62,11 @@ class TestMigrateWithTimeoutsCommand:
                 lock_timeout_in_ms=50_000,
             )
 
-        assert queries[0]["sql"] == "SHOW lock_timeout"
-        assert queries[1]["sql"] == "SET SESSION lock_timeout = '50000ms'"
-        assert queries[2]["sql"] == "SET SESSION lock_timeout = '0'"
-        assert len(queries) == 3
+        assert queries[0]["sql"] == "SELECT pg_try_advisory_lock(8967004653001705610);"
+        assert queries[1]["sql"] == "SHOW lock_timeout"
+        assert queries[2]["sql"] == "SET SESSION lock_timeout = '50000ms'"
+        assert queries[3]["sql"] == "SET SESSION lock_timeout = '0'"
+        assert len(queries) == 4
 
     @mock.patch("django.core.management.commands.migrate.Command.handle", autospec=True)
     @pytest.mark.django_db(transaction=True)
@@ -75,10 +77,11 @@ class TestMigrateWithTimeoutsCommand:
                 statement_timeout_in_ms=50_000,
             )
 
-        assert queries[0]["sql"] == "SHOW statement_timeout"
-        assert queries[1]["sql"] == "SET SESSION statement_timeout = '50000ms'"
-        assert queries[2]["sql"] == "SET SESSION statement_timeout = '0'"
-        assert len(queries) == 3
+        assert queries[0]["sql"] == "SELECT pg_try_advisory_lock(8967004653001705610);"
+        assert queries[1]["sql"] == "SHOW statement_timeout"
+        assert queries[2]["sql"] == "SET SESSION statement_timeout = '50000ms'"
+        assert queries[3]["sql"] == "SET SESSION statement_timeout = '0'"
+        assert len(queries) == 4
 
     def test_interface(self):
         """
@@ -203,6 +206,21 @@ class TestMigrateWithTimeoutsCommand:
                 lock_timeout_in_ms=50_000,
                 lock_timeout_max_retries=2,
                 retry_callback_path=f"{__name__}.example_callback",
+            )
+
+
+class TestLocking:
+    @pytest.mark.django_db(databases=["default", "secondary"])
+    def test_lock_timeout_is_raised(self) -> None:
+        migrate_with_timeouts.Locking().acquire_advisory_session_lock(
+            using="default", value="value"
+        )
+        with pytest.raises(migrate_with_timeouts.LockAlreadyAcquired):
+            migrate_with_timeouts.Locking().acquire_advisory_session_lock(
+                # Secondary is just an alias to "default", so it will
+                # try to acquire the already acquired session-level lock.
+                using="secondary",
+                value="value",
             )
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- The `migrate_with_timeouts` command now acquires a session-level advisory
+  lock before it begins. This helps preventing concurrent instances of the
+  migrate command from running at the same time, which could cause unexpected
+  crashes.
+
 ## [0.1.14] - 2024-12-17
 
 ### Added
--- a/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
+++ b/src/django_pg_migration_tools/management/commands/migrate_with_timeouts.py
@@ -1,5 +1,6 @@
 import dataclasses
 import datetime
+import hashlib
 import importlib
 import io
 import time
@@ -7,6 +8,7 @@ from typing import Any, Protocol, cast
 
 from django.core.management import base
 from django.core.management.commands.migrate import Command as DjangoMigrationMC
+from django.db import connections
 from typing_extensions import Self
 
 from django_pg_migration_tools import timeouts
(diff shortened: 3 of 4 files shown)


tests/django_pg_migration_tools/management/commands/test_migrate_with_timeouts.py
```diff
@@ -206,6 +209,29 @@ def _handle_mock_side_effect(*args: Any, **kwargs: Any) -> None:
             )
 
 
+class TestLocking:
+    @pytest.mark.django_db(transaction=True)
+    @mock.patch(
+        "django_pg_migration_tools.management.commands.migrate_with_timeouts.connections",
+        autospec=True,
+    )
```
marcelofern, line 217: I've ended up opting for a connection mock here.
My original test was using the `threading` module so that I could fire up two
simultaneous connections.

But I had two problems: It slowed tests because one thread had to perform a
call to `time.sleep(1)`, while the other one was firing up. It was a bit flakey
as well depending on how long the other part took to run. I had the same
problems on kraken-core.

For reference, the old code looked like this:

<details>
Old test:

```py

  def test_advisory_lock(self) -> None:
      locked = threading.Event()

      def _acquire_lock() -> None:
          Locking().acquire_advisory_session_lock(...)
          locked.set()
          time.sleep(...)

      sql_thread = threading.Thread(target=_acquire_lock)
      sql_thread.start()

      with pytest.raises(LockAlreadyAcquired):
          Locking().acquire_advisory_session_lock(...)

```
<details>

delfick: personally I'd prefer the tests were a second slower and we didn't use a lock.

Also, it looks like your old test creates a threading event and sets it but doesn't wait for it before trying to make a new advisory session lock?

marcelofern: @delfick Yeah, that's a sketch, I've rewritten a version.
It actually does not need the `time.sleep` so my concerns are covered.

However... Adding threading to the code breaks the test in a different way.
[Pytest doesn't seem happy with psycopg exceptions raised from threaded code](https://docs.pytest.org/en/7.1.x/how-to/failures.html#warning-about-unraisable-exceptions-and-unhandled-thread-exceptions)

<details>

<summary>So for example this code</summary>

This code actually "works" without the `pytest.raises`.
I.e., it raises in the appropriated place.

But it fails with the `pytest.raises` as pytest can't capture
that exception from psycopg anymore.

```python
    @pytest.mark.django_db(transaction=True)
    def test_lock_timeout_is_raised(self) -> None:
        import threading
        locked = threading.Event()

        def _acquire_lock() -> None:
            migrate_with_timeouts.Locking().acquire_advisory_session_lock(
                using="default", value="value"
            )
            locked.set()

        sql_thread = threading.Thread(target=_acquire_lock)
        sql_thread.start()

        locked.wait(timeout=1)
        with pytest.raises(migrate_with_timeouts.LockAlreadyAcquired):
            migrate_with_timeouts.Locking().acquire_advisory_session_lock(
                using="default", value="value"
            )
```

Which fails with:

<details>

<summary>Failure trace</summary>

```sh
======================================== FAILURES =========================================
____________ TestSaferRemoveIndexConcurrently.test_when_not_allowed_to_migrate ____________

cls = <class '_pytest.runner.CallInfo'>
func = <function call_and_report.<locals>.<lambda> at 0x1081d7e20>, when = 'call'
reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)

    @classmethod
    def from_call(
        cls,
        func: Callable[[], TResult],
        when: Literal["collect", "setup", "call", "teardown"],
        reraise: type[BaseException] | tuple[type[BaseException], ...] | None = None,
    ) -> CallInfo[TResult]:
        """Call func, wrapping the result in a CallInfo.

        :param func:
            The function to call. Called without arguments.
        :type func: Callable[[], _pytest.runner.TResult]
        :param when:
            The phase in which the function is called.
        :param reraise:
            Exception or exceptions that shall propagate if raised by the
            function, instead of being wrapped in the CallInfo.
        """
        excinfo = None
        start = timing.time()
        precise_start = timing.perf_counter()
        try:
>           result: TResult | None = func()

../env_django-pg-migration-tools/lib/python3.13/site-packages/_pytest/runner.py:341:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../env_django-pg-migration-tools/lib/python3.13/site-packages/_pytest/runner.py:242: in <lambda>
    lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
../env_django-pg-migration-tools/lib/python3.13/site-packages/pluggy/_hooks.py:513: in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
../env_django-pg-migration-tools/lib/python3.13/site-packages/pluggy/_manager.py:120: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
../env_django-pg-migration-tools/lib/python3.13/site-packages/_pytest/threadexception.py:92: in pytest_runtest_call
    yield from thread_exception_runtest_hook()
../env_django-pg-migration-tools/lib/python3.13/site-packages/_pytest/threadexception.py:68: in thread_exception_runtest_hook
    yield
../env_django-pg-migration-tools/lib/python3.13/site-packages/_pytest/unraisableexception.py:95: in pytest_runtest_call
    yield from unraisable_exception_runtest_hook()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def unraisable_exception_runtest_hook() -> Generator[None]:
        with catch_unraisable_exception() as cm:
            try:
                yield
            finally:
                if cm.unraisable:
                    if cm.unraisable.err_msg is not None:
                        err_msg = cm.unraisable.err_msg
                    else:
                        err_msg = "Exception ignored in"
                    msg = f"{err_msg}: {cm.unraisable.object!r}\n\n"
                    msg += "".join(
                        traceback.format_exception(
                            cm.unraisable.exc_type,
                            cm.unraisable.exc_value,
                            cm.unraisable.exc_traceback,
                        )
                    )
>                   warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
E                   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function BaseConnection.__del__ at 0x107ad7380>
E
E                   Traceback (most recent call last):
E                     File "/Users/marcelo.fernandes/workspace/env_django-pg-migration-tools/lib/python3.13/site-packages/psycopg/_connection_base.py", line 146, in __del__
E                       warn(
E                       ~~~~^
E                           f"connection {self} was deleted while still open."
E                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                           " Please use 'with' or '.close()' to close the connection",
E                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                           ResourceWarning,
E                           ^^^^^^^^^^^^^^^^
E                       )
E                       ^
E                   ResourceWarning: connection <psycopg.Connection [IDLE] (host=localhost port=5416 user=marcelo.fernandes database=test_django_pg_migration_tools) at 0x108240380> was deleted while still open. Please use 'with' or '.close()' to close the connection

../env_django-pg-migration-tools/lib/python3.13/site-packages/_pytest/unraisableexception.py:85: PytestUnraisableExceptionWarning
================================= short test summary info =================================
FAILED tests/django_pg_migration_tools/test_operations.py::TestSaferRemoveIndexConcurrently::test_when_not_allowed_to_migrate - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function BaseConnection.__del__ at 0x107ad7380>

Traceback (most recent call last):
  File "/Users/marcelo.fernandes/workspace/env_django-pg-migration-tools/lib/python3.13/site-packages/psycopg/_connection_base.py", line 146, in __del__
    warn(
    ~~~~^
        f"connection {self} was deleted while still open."
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        " Please use 'with' or '.close()' to close the connection",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        ResourceWarning,
        ^^^^^^^^^^^^^^^^
    )
    ^
ResourceWarning: connection <psycopg.Connection [IDLE] (host=localhost port=5416 user=marcelo.fernandes database=test_django_pg_migration_tools) at 0x108240380> was deleted while still open. Please use 'with' or '.close()' to close the connection
```

</details>

</details>

Closing the connection preemptively in `_acquire_lock` helps, but then the test condition isn't really

marcelofern: Mmm, it works if I add:

```py
    @pytest.mark.filterwarnings(pytest.PytestUnraisableExceptionWarning)
```

So maybe we can go with that. 782c9326cc3c64a6a337b8a9ea798f6715e6fea7

delfick: Let's do this instead

```
class TestLocking:
    @pytest.mark.django_db(transaction=True)
    def test_lock_timeout_is_raised(self) -> None:
        ready_lock=threading.Event()
        done_lock = threading.Event()

        def _acquire_lock() -> None:
            migrate_with_timeouts.Locking().acquire_advisory_session_lock(
                using="default", value="value"
            )
            ready_lock.set()
            done_lock.wait(timeout=1)

        sql_thread = threading.Thread(target=_acquire_lock)
        sql_thread.daemon = True
        sql_thread.start()

        ready_lock.wait(timeout=1)
        with pytest.raises(migrate_with_timeouts.LockAlreadyAcquired):
            migrate_with_timeouts.Locking().acquire_advisory_session_lock(
                using="default", value="value"
            )

        done_lock.set()
        sql_thread.join(timeout=2)


```

I think the error is because you need to unset the lock when you're done with it. Cause the error happens in a different test....

marcelofern: That still doesn't work @delfick. The same error raises.
Even if I add a couple of `clear()` for the events.

It's interesting that it's a different testing failing though,
but I think that might be a sideffect of pytest sharing the
same psycopg connection, which has failed to run in the
LockAlreadyAcquired exception. It should be able to acquire
a new connection for proceding with testing though, that's
probably why this is a warning instead of an error.

Note I have:

```
filterwarnings = "error"
```

On my toml.

delfick: yeah, it fails for me too with that change, but the test itself would be less flakey with it

marcelofern: > Mmm, it works if I add:
> @pytest.mark.filterwarnings(pytest.PytestUnraisableExceptionWarning)

No it doesn't, the test doesn't fail because it doesn't run.

It also doesn't look like the kind of warning I should ignore, considering
a separated test is failing due to the messed up connection state created
by the threads.

---

I found a way better alternative though that doesn't involve threads nor
mocks: c62143aae6399721c9eedf2ba0d0dda45c91b044

delfick: nice

delfick (approved):

ngnpope (approved):

marcelofern:
For digital paper trail: I've tested this in kraken-core and it worked well (-:

PR #65, opened 2025-01-06
Safer Add Field OneToOne

marcelofern:
The default SQL DDL statements produced by Django to create a
OneToOneField are not safe to run and can cause outages.

For example, when the following foo field is added to Bar:
```py
  class Bar(models.Model):
      foo = models.OneToOneField(
          Foo,
          on_delete=models.CASCADE,
          primary_key=False,  # False is the default
          null=True,
      )
```
A new migration is created via makemigrations with the following
operation:
```py
  operations = [
      migrations.AddField(
          model_name='bar',
          name='foo',
          field=models.OneToOneField(
              null=True,
              on_delete=django.db.models.deletion.CASCADE,
              to='blog.foo'
          ),
      ),
  ]
```
This operation, in turn, produces the following statements:
```sql
  BEGIN;
  --
  -- Add field foo to bar
  --
  ALTER TABLE "blog_bar"
  ADD COLUMN "foo_id" bigint NULL
  UNIQUE CONSTRAINT "auto_gen_constraint_name"
  REFERENCES "blog_foo"("id")
  DEFERRABLE INITIALLY DEFERRED;

  SET CONSTRAINTS "auto_gen_constraint_name" IMMEDIATE;
  COMMIT;
```
This is not safe, the ALTER TABLE requires an AccessExclusiveLock on
the bar table, and a ShareRowExclusiveLock on the foo table. Same for
the SET CONSTRAINTS command.

The new SaferAddFieldOneToOne provides an alternative, by creating both
the unique constraint and the fk constraint using the safer operations
that are already included in this library.

### DIFF
--- a/docs/usage/operations.rst
+++ b/docs/usage/operations.rst
@@ -458,7 +458,7 @@ Class Definitions
     **NOTE**: Additional queries triggered by this operation to guarantee
     idempotency have been omitted from the snippet above. The key take away is
     that if this migration fails, it can be attempted again and it will pick up
-    from where it has left.
+    from where it has left off (reentrancy).
 
     How to use
     ----------
@@ -572,7 +572,7 @@ Class Definitions
     **NOTE**: Additional queries that are triggered by this operation to
     guarantee idempotency have been omitted from the snippet above. The key
     take away is that if this migration fails, it can be attempted again and it
-    will pick up from where it has left (reentrancy).
+    will pick up from where it has left off (reentrancy).
 
     **NOTE 2**: If you want to add a ``NOT NULL`` constraint after you have
     backfilled the table, you can use the ``SaferAlterFieldSetNotNull``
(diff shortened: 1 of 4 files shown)


marcelofern:
**Output of a migration generated by this new operation on a sanbox project (with a foo and bar table)**

```
(env) [sandbox] ./manage.py sqlmigrate blog 0020
--
-- Add field foo to bar. Note: Using django_pg_migration_tools SaferAddFieldOneToOne operation.
--

ALTER TABLE "blog_bar"
ADD COLUMN IF NOT EXISTS "foo_id"
bigint NULL;

SET lock_timeout = '0';
CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "bar_foo_id_uniq" ON "blog_bar" ("foo_id");
SET lock_timeout = '0';
ALTER TABLE "blog_bar" ADD CONSTRAINT "bar_foo_id_uniq" UNIQUE USING INDEX "bar_foo_id_uniq";

ALTER TABLE "blog_bar"
ADD CONSTRAINT "blog_bar_foo_id_fk" FOREIGN KEY ("foo_id")
REFERENCES "blog_foo" ("id")
DEFERRABLE INITIALLY DEFERRED
NOT VALID;


ALTER TABLE "blog_bar"
VALIDATE CONSTRAINT "blog_bar_foo_id_fk";


(env) [sandbox] ./manage.py makemigrations && ./manage.py migrate
No changes detected
Operations to perform:
  Apply all migrations: admin, auth, blog, contenttypes, sessions
Running migrations:
  Applying blog.0020_bar_foo... OK
```

src/django_pg_migration_tools/operations.py
```diff
@@ -1303,6 +1327,22 @@ def _maybe_create_index(self) -> None:
                 unique=False,
             )
 
+    def _maybe_create_unique_constraint(self) -> None:
+        if self.unique is True:
```
delfick, line 1331: 🐼 
```suggestion
        if self.unique:
```

delfick (approved):
This seems correct to me, I'd advise waiting for other reviews too though :)

src/django_pg_migration_tools/operations.py
timb07: +1 on this

src/django_pg_migration_tools/operations.py
```diff
@@ -1400,6 +1442,69 @@ def describe(self) -> str:
...
+
+    def __init__(
+        self,
+        model_name: str,
+        name: str,
+        field: models.OneToOneField[models.Model],
+        preserve_default: bool = True,
+    ) -> None:
+        if field.primary_key:
+            raise ValueError("SaferAddFieldOneToOne does not support primary_key=True.")
```
timb07, line 1459: 👍  This is sensible. The only time `primary_key` would be `True` is if the field is being added when the model is added as part of the implementation of multi-table inheritance. And since the DB table doesn't exist yet at that point, there's no need for a "safer" operation.

docs/usage/operations.rst
```diff
@@ -723,3 +723,133 @@ Class Definitions
...
+      -- This operation takes an ShareUpdateExclusiveLock. It won't block
+      -- reads or writes on the table.
+      SET lock_timeout TO '0';
+      CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "bar_foo_id_uniq" ON "myapp_bar" ("foo_id");
+      SET lock_timeout TO '10s';
+
+      -- This operation takes an AccessExclusiveLock, but for a very short
+      -- duration as it leverages the unique constraint index above to create
+      -- the constraint. By default this constraint will be IMMEDIATE, which
+      -- is what we want.
```
timb07, line 785: ❓  Why do we want the constraint to be IMMEDIATE?

docs/usage/operations.rst
```diff
@@ -723,3 +723,133 @@ Class Definitions
...
+
+      -- This query will take a ShareUpdateExclusive lock on the foo table
+      -- (does not block reads nor writes), and a RowShare lock on the bar
+      -- table (does not block reads nor writes).
+      ALTER TABLE foo VALIDATE CONSTRAINT fk_post_bar;
+
+    **NOTE**: Additional queries that are triggered by this operation to
+    guarantee idempotency have been omitted from the snippet above. The key
+    take away is that if this migration fails, it can be attempted again and it
+    will pick up from where it has left (reentrancy).
```
timb07, line 806: typo: "from where it has left **off**"

timb07 (approved):
LGTM

src/django_pg_migration_tools/operations.py
marcelofern: "does `self.unique` quack like a boolean?"

3abf9b567b0759199767bb202837b22c535474c9

docs/usage/operations.rst
marcelofern: ooph that comment is not relevant at all. We can't alter the constraint unless
it's an fk constraint anyway, so it will always be immediate which makes the
comment irrelevant.

```
ALTER CONSTRAINT #

    This form alters the attributes of a constraint that was previously created. Currently only foreign key constraints may be altered.
```

74ab500da23a32d2d5a82b73d11328dd23cea7d0

docs/usage/operations.rst
marcelofern: 6fb244d04d783e449f11ab2297373b64d8c46616

PR #69, opened 2025-01-10
Safer remove field foreign key

marcelofern:
This is the analogous of `SaferAddFieldForeignKey` but does the opposite (-:


s-cooper18 (approved):
LGTM

PR #70, opened 2025-01-10
Fix database_backwards arguments order

marcelofern:
The new_state (from_state) should come before the project_state (to_state), as this is a backwards operation.

There are no changes with regards to how the package works, this is only about test set up.

This change revealed a few problems with the set-up of some tests which have been corrected so that this change could go through.

This set-up has been copied from Django's tests, for example:

https://github.com/django/django/blob/51df0dff3c4f28016185a9e876ee5b3420712f99/tests/migrations/test_operations.py#L308-L314

But something along the way (potentially a wrong ctrl+c/ctrl+v) meant that the order of backward operations remained wrong.

### DIFF
--- a/tests/django_pg_migration_tools/test_operations.py
+++ b/tests/django_pg_migration_tools/test_operations.py
@@ -134,7 +134,7 @@ class TestSaferAddIndexConcurrently:
         with pytest.raises(NotSupportedError):
             with connection.schema_editor(atomic=True) as editor:
                 operation.database_forwards(
-                    self.app_label, editor, project_state, new_state
+                    self.app_label, editor, from_state=project_state, to_state=new_state
                 )
 
     # Disable the overall test transaction because a concurrent index cannot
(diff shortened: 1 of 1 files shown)


tests/django_pg_migration_tools/test_operations.py
```diff
@@ -191,7 +191,7 @@ def test_add(self):
         with connection.schema_editor(atomic=False, collect_sql=False) as editor:
             with utils.CaptureQueriesContext(connection) as queries:
                 operation.database_forwards(
-                    self.app_label, editor, project_state, new_state
+                    self.app_label, editor, new_state, project_state
```
s-cooper18, line 194: Could this be made easier to prevent mistakes in future by requiring the arg names to be specified rather than positional?

s-cooper18: Or even just adding those to your changes just now to make it easier to see in future?

marcelofern: @s-cooper18 let's give it a go: 1eb83fb118964ae2d7a33ada3de8dbbe5d2c3f9d

s-cooper18 (approved):
LGTM

PR #72, opened 2025-01-13
Add deferrable to safer unique constraint

marcelofern:
## Description

Add support for `UniqueConstraint.deferrable`.

Here is a bare Django UniqueConstraint:

```py
class Foo(models.Model):
    int_field = models.IntegerField(null=True)

    class Meta:
        constraints = [
            UniqueConstraint(fields=['int_field'], name='unique_int_field'),
        ]
```

That generates the following SQL:

```sql
ALTER TABLE "blog_foo" ADD CONSTRAINT "unique_int_field" UNIQUE ("int_field");
```

So far so good. But Django allows a `deferrable` argument.
Accepted values are Deferrable.DEFERRED or Deferrable.IMMEDIATE. For example:

```py
from django.db.models import Deferrable, UniqueConstraint

UniqueConstraint(
    fields=['int_field'],
    name='unique_int_field',
    deferrable=Deferrable.DEFERRED,
)
```

If Deferrable.DEFERRED, the migration output is:

```sql
ALTER TABLE "blog_foo" ADD CONSTRAINT "unique_int_field" UNIQUE ("int_field") DEFERRABLE INITIALLY DEFERRED;
```

If Deferrable.IMMEDIATE, the migration output is:

```sql
ALTER TABLE "blog_foo" DROP CONSTRAINT "unique_int_field";
```

The DEFERRABLE INITIALLY IMMEDIATE setup is the default, so Django doesn't have
to alter the DDL statement.

### DIFF
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -455,6 +455,19 @@ class SafeConstraintOperationManager(base_operations.Operation):
         alter_table_sql = base_sql.split(" UNIQUE")[0]
         sql = f'{alter_table_sql} UNIQUE USING INDEX "{index.name}"'
 
+        if constraint.deferrable == models.Deferrable.DEFERRED:
+            sql += " DEFERRABLE INITIALLY DEFERRED"
+        else:
+            # Note that there are only two options for models.Deferrable:
+            # IMMEDIATE or DEFERRED. models.Deferrable.IMMEDIATE is the default
+            # and most common case. DEFERRED is also Postgres default when no
+            # deferrable setting was specified. This means that having an
+            # "else" here that would append "DEFERRABLE INITIALLY IMMEDIATE"
+            # to the query for the "IMMEDIATE" case is not necessary. This
+            # mimics what Django already does, which is to not include that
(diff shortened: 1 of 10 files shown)


src/django_pg_migration_tools/operations.py
```diff
@@ -455,6 +455,19 @@ def create_unique_constraint(
         alter_table_sql = base_sql.split(" UNIQUE")[0]
         sql = f'{alter_table_sql} UNIQUE USING INDEX "{index.name}"'
 
+        if constraint.deferrable == models.Deferrable.DEFERRED:
+            sql += " DEFERRABLE INITIALLY DEFERRED"
+        else:
+            # Note that there are only too options for models.Deferrable:
```
jamesbeith, line 461:
```suggestion
            # Note that there are only two options for models.Deferrable:
```

jamesbeith (approved):

src/django_pg_migration_tools/operations.py
marcelofern: Classic 😄 e3a68061b9cd79492e6c608b0a5df72bb6d8b94a

PR #75, opened 2025-02-10
Fix safer alter field set not null for fk fields

marcelofern:
Changes:

    Fix SaferAlterFieldSetNotNull for custom-named fields
    
    Prior to this change, this operation didn't work for fields that had
    column names in the database that didn't match their names (such as
    ForeignKey fields, which have an "_id" suffix).
    
    This commit changes that by obtaining the column name from the model's
    meta class instead.

### DIFF
--- a/tests/django_pg_migration_tools/test_operations.py
+++ b/tests/django_pg_migration_tools/test_operations.py
@@ -1693,7 +1693,6 @@ class TestSaferAlterFieldSetNotNull:
         # the router.
         assert len(queries) == 0
 
-    @pytest.mark.xfail
     @pytest.mark.django_db(transaction=True)
     def test_when_field_is_a_foreign_key(self):
         project_state = ProjectState()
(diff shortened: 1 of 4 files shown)


tests/django_pg_migration_tools/test_operations.py
```diff
@@ -1692,8 +1693,102 @@ def test_when_not_allowed_to_migrate_by_the_router(self):
...
+            SELECT 1
+            FROM pg_catalog.pg_attribute
+            WHERE
+                attrelid = 'example_app_nullfkfieldmodel'::regclass
+                AND attname = 'fk_id'
+                AND attnotnull IS TRUE;
+        """)
+
+    @pytest.mark.django_db(transaction=True)
+    def test_when_field_is_a_foreign_key(self):
```
timb07, line 1790: I think this test name got switched with the one for the new test above.

marcelofern: It did! Will amend.

PR #77, opened 2025-02-20
Add documentation guides and background

marcelofern:
# Description

Much of our documentation is nested inside the "class definition" of each
operation.

The problem with that is that it is not obvious to a user coming to the repo
to know that there will be tutorials inside the class definition documentation.

This PR pulls out these tutorials/guides into a separated documentation page.

The content does not change, with the exception of a few adjustments when
necessary to fit the new `.rst` file structure and some double-linked links to
go from guide to class definition and vice-versa.

The new nav-bar looks like this:

image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,6 +18,49 @@ Install from pypi::
 
     python -m pip install django-pg-migration-tools
 
+Background
+==========
+
+In the past, traditional applications would do schema changes during scheduled
+"maintenance windows." During this time, the application would stop all
+traffic, and developers could make changes to the database or servers without
+worrying about users.
+
+In the context of modern systems requiring 24/7 availability, this is no longer
+possible.
+
+Django's migration system usually picks the easiest way to make a schema
+change, but that isn't always safe. In fact, some operations Django does can
+cause problems, like locking tables for too long, which can cause outages.
+
+.. dropdown:: Meaning of **"safe"** in this context.
+    :color: info
+    :icon: info
+
+    "Safe" means that the migration operation can run during normal application
+    operating hours without interrupting the availability of the database or
+    causing errors while using the database.
+
+Modern applications also tend to make schema changes before deploying new code.
+For example, in a blue/green deployment, a schema change is made to the
+database, and then traffic is slowly moved from old servers (blue) to new
+servers (green).
+
+Once all traffic is migrated to the "green" servers, the "blue" servers usually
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -18,6 +18,40 @@ Install from pypi::
...
+In the past, traditional applications would do schema changes during scheduled
+"maintenance windows." During this time, the application would stop all
+traffic, and developers could make changes to the database or servers without
+worrying about users.
+
+In the context of modern systems requiring 24/7 availability, this is no longer
+possible.
+
+Django's migration system usually picks the easiest way to make a schema
+change, but that isn't always safe. In fact, some operations Django does can
```
timb07, line 33: It would be good to define "safe" here, particularly since many operations are named "…Safer…". The rest of the "Background" section explains what the implications and requirements of being safe are, but it would be good to have a short summary first. E.g.

> By "safe", we mean that the migration operation can run during normal application operating hours without interrupting the availability of the database or causing errors while using the database.

docs/index.rst
```diff
@@ -18,6 +18,40 @@ Install from pypi::
...
+cause problems, like locking tables for too long, which can cause outages.
+
+Modern applications also tend to make schema changes before deploying new code.
+For example, in a blue/green deployment, traffic is slowly moved from old
+servers (blue) to new servers (green).
+
+Once all traffic is migrated to the "green" servers, the "blue" servers usually
+remain ready in case the application needs to be rolled back.
```
timb07, line 41: This examnple is missing information about when the schema changes are made.

Suggestion:

> For example, in a blue/green deployment, a schema change is made to the database, and then traffic is slowly moved from old servers (blue) to new servers (green).

timb07 (approved):
Looking good.

I've made some suggestions re the new "Background" section.

Since the other docs have mostly just been moved, I haven't re-read them to check for errors. I did quick checks to see that the links left behind in the code seemed to be correct (but I didn't pull down the branch to actually verify them).

docs/index.rst
marcelofern: Good idea. I have added an info component here: 6d6763313db93e0452903c00a7b8ff0a9da78893

docs/index.rst
marcelofern: Good idea. Changed here: f5ace49c67facbaeb4971868cbcbc1f7b5423cac

PR #79, opened 2025-04-02
Fix `SaferAddFieldForeignKey` when field name != column name

timb07:
Prior to this change, this operation didn't work for fields that referred to tables where the primary key field had a column name that differed from the field name.

This commit fixes that by referring to the column name instead of the field name.


marcelofern (approved):
Thanks for the fix. I am not sure why the coverage bot is broken, one more
thing to the #TODO list. I have checked locally though, and the coverage is
still 100% (as is obvious by the diff).

PR #84, opened 2025-06-17
Bugfix: drop unique constraint sqlmigrate is noop

s-cooper18:
Change `collect_default` on constraint existing checks to require an explicit value rather than defaulting to false. This was causing the operation to say that the constraint did not exist and could not generate any SQL when the migration to drop a constraint was in collect only mode.

The value needs to be set to True when dropping constraints so that the SQL for what would happen if a constraint exists is shown (as the alternative is not useful).

### DIFF
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -506,7 +506,7 @@ class SafeConstraintOperationManager(base_operations.Operation):
             )
             return
 
-        if not self._constraint_exists(schema_editor, constraint):
+        if not self._constraint_exists(schema_editor, constraint, collect_default=True):
             # Nothing to delete.
             return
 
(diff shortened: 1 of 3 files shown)


src/django_pg_migration_tools/operations.py
```diff
@@ -506,7 +506,7 @@ def drop_unique_constraint(
             )
             return
 
-        if not self._constraint_exists(schema_editor, constraint):
+        if not self._constraint_exists(schema_editor, constraint, collect_default=True):
```
flpcury, line 509: Just to check if I understood it:
For `sqlmigrate` mode, there's no real introspection. So this bit of logic would fake the introspection to always say that the constraint doesn't exist. Now it will always say that the constraint does exist.
Since there's no introspection, neither approach is guaranteed to produce the actual SQL that would be produced when running it for real. But the new approach of defaulting to "there is a constraint" is more informative (it better reflects what the user is trying to accomplish with the operation)

flpcury (approved):
Nice find!

src/django_pg_migration_tools/operations.py
s-cooper18: Pretty much

PR #89, opened 2025-08-07
Fix SaferRemoveFieldForeignKey for non null columns

marcelofern:
Prior to this change, if the FK field being removed had a null=False set up, this operation would fail:

  ValueError: Can't safely create a FK field with null=False

While this error is relevant when the FK manager is *creating* a new column, it is not relevant when deleting one.

This commit changes that by skipping the nullability check when running from SaferRemoveFieldForeignKey.

A note was dropped on the documentation to alert that the field will be created with null=True when reversing the operation, since the alternative wouldn't be safe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


flpcury (approved):
Nice 👍

PR #99, opened 2026-02-25
Add support for `nulls_distinct=False` in unique constraints

timb07:
A migration with the `SaferAddUniqueConstraint` operation for a constraint using `nulls_distinct=False` failed with a syntax error when attempted against Postgres version 14.x:

```
syntax error at or near "None"
LINE 1: ...mjobs/management/commands/migrate.py', line=28 */ None UNIQU...
                                                             ^
```

The ultimate cause was attempting to create a unique constraint with "NULLS NOT DISTINCT", which isn't supported on Postgres until version 15. And in the course of investigating that issue, we also noticed that even on Postgres 15 and higher, the constraint is created, but without "NULLS NOT DISTINCT".

This PR:

- adds support for creating unique constraints with `nulls_distinct=False`
- adds better error handling for when attempting to create such constraints on Postgres versions that don't support them (version 14 and earlier)

Note that this PR also introduces 3 missed lines in the code coverage of test_operations.py; it's unavoidable since code coverage is run with Postgres 16, and those lines are only exercised with Postgres 14 and below.

### DIFF
--- a/src/django_pg_migration_tools/operations.py
+++ b/src/django_pg_migration_tools/operations.py
@@ -267,6 +267,7 @@ class SafeIndexOperationManager(
         index: models.Index | IndexSQLBuilder,
         unique: bool,
         model: type[models.Model],
+        nulls_distinct: bool = True,
     ) -> None:
         self._ensure_not_in_transaction(schema_editor)
 
@@ -283,6 +284,7 @@ class SafeIndexOperationManager(
             model=model,
             schema_editor=schema_editor,
             index=index,
+            nulls_distinct=nulls_distinct,
         )
         schema_editor.execute(index_sql)
 
@@ -374,6 +376,7 @@ class SafeIndexOperationManager(
         model: type[models.Model],
         schema_editor: base_schema.BaseDatabaseSchemaEditor,
         index: models.Index | IndexSQLBuilder,
+        nulls_distinct: bool,
     ) -> str:
         if isinstance(index, IndexSQLBuilder):
             assert model._meta.db_table == index.table_name
@@ -388,6 +391,9 @@ class SafeIndexOperationManager(
         if unique:
             index_sql = index_sql.replace("CREATE INDEX", "CREATE UNIQUE INDEX")
 
+        if not nulls_distinct:
+            index_sql += " NULLS NOT DISTINCT"
+
         return index_sql
 
 
@@ -396,6 +402,10 @@ class ConstraintOperationError(Exception):
 
 
 class ConstraintAlreadyExists(ConstraintOperationError):
+    pass
+
+
+class ConstraintNotSupported(ConstraintOperationError):
     pass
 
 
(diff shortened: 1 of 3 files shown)


src/django_pg_migration_tools/operations.py
```diff
@@ -388,6 +391,9 @@ def _get_create_index_sql(
         if unique:
             index_sql = index_sql.replace("CREATE INDEX", "CREATE UNIQUE INDEX")
 
+        if nulls_distinct is not None and not nulls_distinct:
+            index_sql += " NULLS NOT DISTINCT"
```
s-cooper18, line 395: ❓ So the default behaviour is None, which is the same as `nulls_distinct=True`?

Does it make more sense to have it be a boolean True or False instead with a default of True?

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -1436,6 +1437,144 @@ def test_when_expression_on_constraint_only_creates_index(self):
...
+                name="unique_null_int_field",
+                nulls_distinct=False,
+            ),
+        )
+        operation.state_forwards(self.app_label, new_state)
+        # Proceed to add the unique index followed by the constraint:
+        with connection.schema_editor(atomic=False, collect_sql=False) as editor:
+            if not connection.features.supports_nulls_distinct_unique_constraints:
+                # Postgres versions <= 14 do not support NULLS NOT DISTINCT
+                with pytest.raises(operations.ConstraintNotSupported):
```
s-cooper18, line 1481: The test is structured in this way because the entire test suite is run across multiple versions of Postgres?

tests/django_pg_migration_tools/test_operations.py
```diff
@@ -1436,6 +1437,144 @@ def test_when_expression_on_constraint_only_creates_index(self):
...
+            model_name="nullintfieldmodel",
+            constraint=UniqueConstraint(
+                fields=("int_field",),
+                name="unique_null_int_field",
+                nulls_distinct=False,
+            ),
+        )
+        with connection.schema_editor(atomic=False, collect_sql=False) as editor:
+            with patch.object(
+                connection, "features", supports_nulls_distinct_unique_constraints=False
```
s-cooper18, line 1567: ❓ You have a distinct test for the scenario where not supported - should the other test be changed to only be testing the scenario where it is in fact supported? That is, patching the it is supported part in the previous test as well?

tests/django_pg_migration_tools/test_operations.py
timb07: Yes, that's right. But code coverage is only run with one version of Postgres, so it complains about the above.

tests/django_pg_migration_tools/test_operations.py
timb07: This test was added to ensure complete code coverage when running with Postgres 16 (which is what coverage runs with).

The other test doesn't use patching, so it's a better test that the operation behaves as expected with each supported Postgres version.

src/django_pg_migration_tools/operations.py
timb07: > So the default behaviour is None, which is the same as `nulls_distinct=True`?

I just checked what Django does. With `nulls_distinct` it appends `NULLS DISTINCT` to the SQL. That's functionally the same as omitting any reference to NULLs.

Rather than arguing whether to change to only supporting `True` and `False`, I think it's better to align the behaviour here with what Django does.

timb07: Update: I've changed my mind. There's no value in having three distinct options for `nulls_distinct`, and it will only add complexity and the need for further tests. So we'll diverge from Django behaviour, but with the benefit of some more clarity and simpler logic in a few places.

timb07: Fixup: 11cef5412fbcbea6210cf66ddd96cf30fa05440f

s-cooper18 (approved):
Thanks for the changes - looks good to me 👍

